PR #1, opened 2021-05-18
Fix Winit link in README by adding missing https

SebastianAigner:
Original link in README did not specify protocol, was falsely treated as a relative link.


tonsky:
Oops, thank you!

Issue #7, opened 2021-07-01
Create CLion project


tonsky:
- [ ] Investigate placing CMakeLists.txt to the root folder

PR #17, opened 2021-07-02
Setup basic root cmake file (for clion project import)

EgorOrachyov:
#7  It looks kinda weired for now since it has no complex logic, mb in the future we will extend it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@ target
 build
 __pycache__
 .idea
+
+# clion build directories
+/*build*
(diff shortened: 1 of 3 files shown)


EgorOrachyov:
@tonsky @SergeevPavel what do you think?

EgorOrachyov:
CMAKE_SOURCE_DIR won't work in macos end etc., if add this folders as subdirs, since
when root cmake is ...../JWM/CMakeLists.txt => CMAKE_SOURCE_DIR  = ..../JWM, and not ..../JWM/macos

EgorOrachyov:
I'll replace CMAKE_SOURCE_DIR  with CMAKE_CURRENT_LIST_DIR (must work exactly the same way)

SergeevPavel:
works for me! thanks

PR #83, opened 2021-08-04
Prototype clipboard api

EgorOrachyov:


shared/java/org/jetbrains/jwm/ClipboardEntry.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @NotNull @SneakyThrows
+    public static ClipboardEntry makeString(ClipboardFormat format, String text) {
+        return make(format, text.getBytes("UTF-16LE"));
+    }
+
+    @NotNull @SneakyThrows
+    public String getString() {
+        assert ClipboardFormat.isStringCompatible(_format);
```
tonsky, line 38: What if user register her own string-compatible format?

shared/java/org/jetbrains/jwm/Clipboard.java
```diff
@@ -0,0 +1,58 @@
...
+            return format;
+
+        if (!_nRegisterFormat(formatId))
+            throw new RuntimeException("Failed to register format: " + formatId);
+
+        return registerFormatInternal(formatId);
+    }
+
+    @ApiStatus.Internal @NotNull
+    public static ClipboardFormat registerFormatInternal(String formatId) {
```
tonsky, line 47: Our convention is to use `_` to indicate internal methods/fields.

shared/java/org/jetbrains/jwm/Clipboard.java
```diff
@@ -0,0 +1,58 @@
...
+    public static ClipboardFormat registerFormat(String formatId) {
+        ClipboardFormat format = _formats.get(formatId);
+
+        if (format != null)
+            return format;
+
+        if (!_nRegisterFormat(formatId))
+            throw new RuntimeException("Failed to register format: " + formatId);
+
+        return registerFormatInternal(formatId);
```
tonsky, line 43: Need some lock between checking if format exists and actually registering new format

tonsky (changes requested):
Looks great, some minor details need polishing (see comments)

shared/java/org/jetbrains/jwm/Clipboard.java
EgorOrachyov: I thought, our api calls are allowed only on UI thread?

PR #124, opened 2021-08-22
add a detailed Windows build document

ItoYo16u:
Hello, and thank you for your great work!

I added a document about build on Windows.

Besides, I added a note for running the example application since it seems that the example application requires JDK 16


tonsky:
Thank you!

Issue #99, opened 2021-08-16
Set system cursor


tonsky:
enum MouseCursor
Window::setMouseCursor(MouseCursor)

- [ ] Windows
- [ ] Linux
- [ ] macOS

PR #127, opened 2021-08-23
Windows/Shared: Set system mouse cursor #99

EgorOrachyov:


shared/java/org/jetbrains/jwm/MouseCursor.java
```diff
@@ -0,0 +1,18 @@
...
+
+public enum MouseCursor {
+    ARROW  (0),
+    CROSS  (1),
+    HAND   (2),
+    HELP   (3),
+    IBEAM  (4),
+    UPARROW(5);
+
+    @ApiStatus.Internal public final int _id;
```
tonsky, line 13: Let’s use `_values` to transform enum <-> int, just for consistency with Key.java

shared/java/org/jetbrains/jwm/MouseCursor.java
```diff
@@ -0,0 +1,18 @@
+package org.jetbrains.jwm;
+
+import org.jetbrains.annotations.*;
+
+public enum MouseCursor {
+    ARROW  (0),
+    CROSS  (1),
+    HAND   (2),
```
tonsky, line 8: macOS has three types of hands: open, closed and pointing. Let’s have more specific name here? E.g. HAND_POINTING, HAND_OPEN or HAND_CLOSED?

shared/java/org/jetbrains/jwm/MouseCursor.java
```diff
@@ -0,0 +1,18 @@
+package org.jetbrains.jwm;
+
+import org.jetbrains.annotations.*;
+
+public enum MouseCursor {
+    ARROW  (0),
+    CROSS  (1),
```
tonsky, line 7: How about CROSSHAIR instead of CROSS?

tonsky (changes requested):
Looks good, few minor naming suggestions

shared/java/org/jetbrains/jwm/MouseCursor.java
EgorOrachyov: ID used to transform java enum to c++. 
There is no need to transform c++ to java => no values needed

shared/java/org/jetbrains/jwm/MouseCursor.java
EgorOrachyov: Ok, but on windows they all will be mapped to IDC_HAND

tonsky: Let's leave only one, the one that Windows has?

shared/java/org/jetbrains/jwm/MouseCursor.java
tonsky: IBEAM.ordinal() will get you the integer

EgorOrachyov: 👌

shared/java/org/jetbrains/jwm/MouseCursor.java
EgorOrachyov: Windows has IDC_HAND

EgorOrachyov: Or we can do something like that in win

image

PR #134, opened 2021-08-25
Feature: window opacity

ItoYo16u:
add set opacity function

- [x] setOpacity
- [ ] getOpacity


shared/java/org/jetbrains/jwm/Window.java
```diff
@@ -57,6 +57,12 @@ public UIRect getContentRectAbsolute() {
     @NotNull @Contract("-> this")
     public abstract Window setIcon(File icon);
     public abstract Window setMouseCursor(MouseCursor cursor);
+    
+    /**
+    *  Sets window opacity (0 - 255)
+     */
+    @NotNull @Contract("-> this")
+    public abstract Window setOpacity(int opacity);
```
tonsky, line 65: - I want to propose float 0.0-1.0 instead of int value, to match Electron and macOS APIs
- Add getOpactiy method to complement setOpacity

windows/cc/WindowWin32.cc
```diff
@@ -74,6 +74,15 @@ void jwm::WindowWin32::setIcon(const std::wstring& iconPath) {
     HICON hicon = (HICON)LoadImage(NULL, iconPath.c_str(), IMAGE_ICON, 0, 0, LR_LOADFROMFILE | LR_DEFAULTSIZE);
     SendMessage(_hWnd, WM_SETICON, ICON_SMALL, (LPARAM)hicon);
 }
+
+void jwm::WindowWin32::setOpacity(int opacity) {
+    JWM_VERBOSE("Set window opacity'" << opacity << "'");
+    LONG previousStyle = GetWindowLong(_hWnd, GWL_EXSTYLE);
+    LONG opacityEnabled = previousStyle |= WS_EX_LAYERED;
+    SetWindowLong(_hWnd, GWL_EXSTYLE, opacityEnabled);
```
tonsky, line 82: I am not proficient in WinAPI, but it feels like if opacity is set to 1.0 (fully opaque), we should remove WS_EX_LAYERED instead of adding it. Correct me if I’m wrong.

tonsky (changes requested):
Thanks again for your contribution! I added three minor suggestions, otherwise looks very good!

shared/java/org/jetbrains/jwm/Window.java
ItoYo16u: Ok. I'll change it.

windows/cc/WindowWin32.cc
ItoYo16u: Yes, I think it is better. I will fix it.

tonsky:
Thanks, looks great! Merged

tonsky:
Sorry, I missed the fact that you didn’t get time to implement getOpacity. Send in another PR?

PR #143, opened 2021-08-28
[wip] Windows: system theme

i10416:
- [x] get system theme
- [x] disable when high contrast mode
- [x]  set theme 
- [x]  expose isHighContrast

Add functions to get window theme and switch light/dark mode. 

I implemented it by referring to [winit]( https://github.com/rust-windowing/winit/blob/master/src/platform_impl/windows/dark_mode.rs) and [this](https://gist.github.com/ysc3839/b08d2bff1c7dacde529bed1d37e85ccf).

See also:
- https://github.com/ysc3839/win32-darkmode


tonsky:
Hi!

Thanks again for taking time and effort to tackle this. This is an important issue we’d love to see it merged in JWM as soon as possible.

I know your branch is still WIP, but @EgorOrachyov and I took a quick look and have a bit of feedback to share so far. I hope you don’t mind.

1. Early on, we decided to only target Windows 10+, mostly for implementation simplicity and ease of testing. If we only target Windows 10+, it feels that dynamic version check (user32.dll) and uxtheme.dll are excessive. @EgorOrachyov thinks linking uxtheme.dll in CMakeLists.txt is enough, what do you think?

2. Dark/Light are not the only flags that modern OSes use to control apps’s look. Another two are High Contrast and Inverted colors. Do you think you can accomodate that in the API? One option might be to make Theme a data class with three booleans: `isDark`, `isHighContrast`, `isInverted`? This Electron seems to be [doing the same](https://www.electronjs.org/docs/api/native-theme)

3. This is minor, but we are using 4 spaces indentation in C++ and `{` on the same line. It would be great for consistency if you switched to that style also.

Thanks again and glad to see you among contributors!

ItoYo16u:
@tonsky 
Thank you for helpful feedback!
> it feels that dynamic version check (user32.dll) and uxtheme.dll are excessive.

Ok, I will separate function respectively and use only uxtheme.dll part for the time being. Or, should I drop current implementation?

> Dark/Light are not the only flags that modern OSes use to control apps’s look. 

I agree with you. It would be better. I will check Electron implementation and fix my code.
 
> This is minor, but we are using 4 spaces indentation in C++ and { on the same line. 

Sorry 🙇

ItoYo16u:
- https://github.com/electron/electron/blob/c0e72bd3357c1edaff30087307690d4a113bd01a/docs/tutorial/dark-mode.md
- https://github.com/electron/electron/blob/44491b023ac2653538b7ac36bb73f3085a938666/shell/browser/win/dark_mode.cc
- https://github.com/electron/electron/blob/44491b023ac2653538b7ac36bb73f3085a938666/shell/browser/api/electron_api_system_preferences.cc

shared/java/org/jetbrains/jwm/AppearancePreference.scala
```diff
@@ -0,0 +1,12 @@
...
+import org.jetbrains.annotations.*;
+import lombok.*;
+
+
+@Data
+public class AppearancePreference {
+    private final Theme theme;
+    private final boolean isHighConstast;
+    private final boolean isInverted;
+}
```
ItoYo16u, line 12:
```suggestion
} 

```

ItoYo16u:
@tonsky 
After investigating Windows API and Electron implementation, I found that Windows does not offer  API to check whether or not inverted color mode is enabled. Besides, it seems electron uses chromium properties to determine InvertedColorScheme mode.  

Perhaps I may miss something, but I think `isInverted` is not as much relevant as `isHighContrast` since Windows disables dark mode when it is in high contrast mode whereas it does not when in inverted color mode.
Thus, could I leave isInverted out  for this PR?

tonsky:
> Thus, could I leave isInverted out for this PR?

Sure, we can come back to it later

tonsky:
From what I see on Electron, not all of these properties are supported on all platforms. It might be that Windows doesn’t even have inverted mode (does it?)

ItoYo16u:
As far as I konw, Windows does have invert color scheme(ctrl+alt+I), but  I could not find a way to programatically detect/control it.(There is InvertRect api, but it does not suit for use case.   )
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-invertrect

I also have a quick look at chromium implementation.  

https://chromium.googlesource.com/chromium/src/+/refs/heads/main/ui/native_theme/native_theme_win.cc

tonsky:
I’m not sure how is Github supposed to communicate when this is ready to review, so please let me know!

So far, I took a quick look and have a few questions:

- How do you see `AppearancePreference` might be used?
- In which cases `Theme::SYSTEM` is used?

i10416:
@tonsky  I think this PR becomes a bit messy, so I will separate changes in other PRs.

PR #153, opened 2021-09-19
[Windows] add is highcontrast

i10416:
separate changes in #143

I separated #143 because it becomes a bit messy.
This PR contains changes related to `isHighcontrast` functionality.


tonsky:
Thanks! I'll take a look next week

shared/java/org/jetbrains/jwm/Window.java
```diff
@@ -232,6 +232,11 @@ public Window setMouseCursor(MouseCursor cursor) {
      */
     public abstract float getOpacity();
 
+    /**
+     * <p>check whether OS currently uses high contrast mode.</p>
+     */
+    public abstract boolean isHighContrast();
```
tonsky, line 238: You were correct in #143 that themes are per-app, not per-window. Apologies if it was me who confused and/or suggested this method to be on a Window class.

It could either go on App or on a new class called Theme, similar how Clipboard groups all the functions related to clipboard manipulation. I suggest Theme.

Note: this would be different from dark/light enum, this will be a class containing static methods like isHighContrast() etc. We’ll have to come up with something different re: dark/light

shared/cc/ThemeHelper.hh
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+namespace jwm {
+    class ThemeHelper {
```
tonsky, line 4: Could be renamed to Theme, to match Java

windows/cc/ThemeHelperWin32.hh
```diff
@@ -0,0 +1,20 @@
...
+#include <PlatformWin32.hh>
+
+#include "ThemeHelper.hh"
+#include "ThemeHelperWin32.hh"
+
+namespace jwm {
+  
+    class ThemeHelperWin32 final : public ThemeHelper {
+       public:
+           ThemeHelperWin32(HWND hWnd) { _hWnd = hWnd; };
```
tonsky, line 11: HWND seems unused

tonsky (changes requested):
Great idea to split PR into small chunks! I like the implementation, the only issue I have with it is that it should be on a separate static class, now on a Window (apologies if I ever suggested otherwise)

shared/java/org/jetbrains/jwm/Window.java
i10416: According to your advice, I added a Theme class.

i10416:
I have one question. (I am not quite familiar with GUI app and  this may be a stupid question).
`_onUIThread` is called everywhere, what is this for?

tonsky:
> I have one question. (I am not quite familiar with GUI app and this may be a stupid question).
> `_onUIThread` is called everywhere, what is this for?

It’s important that all communication with OS APIs, either it’s on Windows, Linux or Mac, happen from the same thread on which window was created. We (and most of the other UI framewors) call it UI thread. It’s usually a single dedicated thread where even loop was started and it’s usually blocked in that loop.

`_onUIThread` is an assertion that you are, actually, on that thread when the method is called. If you are not, you get and exception and should fix it.

Simple rule of thumb would be to put it everywhere.

tonsky:
Thanks, merged

tonsky:
I’ve pushed macOS impl and removed C++ ThemeWin32 class, seems unnecessary. Please take a look https://github.com/HumbleUI/JWM/commit/5a680f37cd91b31bc4c19d6d1349a707be917bc5

tonsky:
Actually, commit is now https://github.com/HumbleUI/JWM/commit/bada6137e05960081014b1d015dd093200fee65e, I rebased to fix compilation

PR #159, opened 2021-09-28
[DO NOT MERGE] Initial script to build a working Graal binary, Windows

GavinRay97:
I deleted most of the example stuff just to get this working without building Skija locally since the dependency/build mgmt is done through Python scripts instead of Gradle/Maven (felt easier lol)

So don't merge this -- but the changes to `native_image.py` show a working (for me) path towards building a Windows binary. You could build a shared/static lib with `@CEntrypoint()` and consume the Java app from C/C++ too.

I am able to implement and test this for Linux as well, but I cannot test for Darwin unfortunately, as I don't own an Apple computer.

The Graal configuration needed for JNI and resources (the shared lib) to work can be distributed by putting them in the `.jar`, under: `META-INF/native-image/<group-id>/<artifact-id>`, so I've done that here with the output of the tracing agent:

image

Note there are some rough spots yet. Specifically:
```java
    public static void main(String[] args) {
        App.init();

        // Window window = App.makeWindow();

        // Using the generic window seems to cause:
        // Exception in thread "main" java.lang.NoSuchMethodException: io.github.humbleui.jwm.WindowWin32.<init>()
        //     at java.lang.Class.getConstructor0(DynamicHub.java:3585)
        //     at java.lang.Class.getDeclaredConstructor(DynamicHub.java:2754)
        //     at io.github.humbleui.jwm.App.makeWindow(App.java:50)
        //     at io.github.humbleui.jwm.examples.Example.main(Example.java:18)

        // Maybe because the underlying constructor isn't directly invoked, so Graal doesn't statically analyze it?
        // Potential solution: Hard-code "if OS == WINDOWS { new WindowWin32() }" etc?

        WindowWin32 window = new WindowWin32();
```

I'm no Graal expert, just someone who has used it a bit. Can try to figure out more about this, but I think it may be because it never encounters the direct constructor invocation so doesn't include it in the compiled binary?

Not really sure.


tonsky:
This is fantastic! I’ll take a look soon and see if I can run on/adopt it to macOS. Being able to run from native image is a very desirable property!

tonsky:
The exception you see is probably due to reflection in App::makeWindow. Removed in https://github.com/HumbleUI/JWM/commit/2a02339790fcfaf200dd6c33df6ffab7d7055abd

GavinRay97:
Ohh, 🤦 yeah that makes complete sense.
I'd bet money you're spot-on with that change, would be surprised if it didn't fix it tbh.

PR #177, opened 2021-12-30
Update Getting Started maven dependency version

Nithanim:
Update the maven dependency version in the "Getting Started" docs to 0.3.1 shown in in https://search.maven.org/search?q=a:jwm

### DIFF
--- a/docs/Getting Started.md
+++ b/docs/Getting Started.md
@@ -16,7 +16,7 @@ E.g. for Maven it’ll look like this:
 <dependency>
     <groupId>io.github.humbleui</groupId>
     <artifactId>jwm</artifactId>
-    <version>0.2.6</version>
+    <version>0.3.1</version>
 </dependency>
 ```
 


tonsky:
I think I’ll need to update Getting Started as well, the versions are not compatible

Nithanim:
Yeah, you are right. I did not get too far yesterday. 

To be completely honest, the "Getting Started" experience I am having is a bit rough. The given maven dependency is wrong (as fixed in this PR) and then there are compile errors in the scaling examples. But I am not even sure what the "specific to your platform" in the [first sentence](https://github.com/HumbleUI/JWM/blob/main/docs/Getting%20Started.md?plain=1#L5) is supposed to mean because there is only a single dependency given?

Additionally the [last sentence about the dashboard skija integration example](https://github.com/HumbleUI/JWM/blob/main/docs/Getting%20Started.md?plain=1#L260) does not really help. No file in the linked folder matches the given name. Also, all imports are *-imports so you have no idea where the used classes come from (skija or JWM). This makes it hard to differentiate if you have no experience in any of those libraries.

tonsky:
Sorry about that. I just finished a huge refactoring and didn’t updated Getting Started yet. Will do after the holidays

Nithanim:
No problem! Thank you very much! :smiley:

tonsky:
Updated Getting Started. Let me know if something could be improved further

Nithanim:
Thank you very much!
`window.setWindowSize(int)` is still passed a float value. Otherwise it looks good now.

Issue #75, opened 2021-07-30
API: Customize window titlebar


tonsky:
https://web.dev/window-controls-overlay/

mworzala:
Has any further thought been put into this API? I am interested in implementing it for macOS.

Does this include macOS titlebars with toolbars? or full customization of traffic light position like electron has?

Is there any plan to solve the issue of drag regions without a titlebar?

tonsky:
That’s great! It’s a great and important feature for sure.

I only have general idea of how this should look like. I think the main use case here is to do title-less windows. It would require to position traffic lights / windows close buttons and drag regions, as I imagined. This is as far as “cross-platform” part goes. (it could be implemented partially, e.g. only for macOS first, but keeping cross-platform usage in mind).

There could also be macOS APIs that do macOS-specific things, I think (as methods on WindowMac, for example).

I am happy to review API proposal before implementation.

Alex2772:
macOS has the simplest way to implement this - you could place any content in the titlebar and offset "traffic lights" to any place you want.

On Windows, you should draw and handle the whole titlebar by yourself (you can notice the Windows 10-styled window control buttons when running modern software with custom titlebar (even Microsoft Office) on Windows 7 or 8). I've tried to do the job and I think it's not hard.
![image](https://user-images.githubusercontent.com/19491414/148299703-7277dcc0-4647-4cc4-b6de-d1e6bc23fbb5.png)

On Linux of course the situation is the worst: there're lots of desktop environments. You can possibly to do the job with GTK (there're lots of GTK applications with customized titlebar) but what about KDE users? What about DEs without titlebars (i.e. i3wm)? Looks like the best way on Linux is to turn off custom titlebar because you could not guarantee how does a custom titlebar would look like and even would it exist.

Goods news: all three platforms support custom drag regions.

mworzala:
Drawing a similarly styled titlebar seems out of scope of the project to me, but correct me if I am wrong.

In my mind there would be a way to hide the titlebar and add drag regions on all (or as many as feasible re linux) platforms, and then some platform specific functions for things like positioning the traffic lights on macOS in their respective window classes. 

Will come back with an API proposal to review once I work out how drag regions work :)

tonsky:
Sounds good!

mworzala:
**Titlebar Settings**

Based on my understanding, every platform window can have its titlebar removed completely (as follows on macOS).
[...]

PR #181, opened 2022-01-11
macOS titlebar options

mworzala:
Implements titlebar customization as outlined in #75 for macOS.

Missing user drag regions
Missing some javadocs

Some behaviors:
- I am still able to drag the window near the top when using a full size content area.
- This is using a different method than Electron for custom traffic light position. This method seems more reliable, however it does not allow the buttons to exit the titlebar frame. I have not checked if electron does, but if so and its desired behavior I may need to switch to electrons method.
- Currently traffic light positioning is ignored if the titlebar is hidden. It is because the underlying buttons are gone when titlebar is removed. I assume this could be worked around (by recreating the buttons), but I am not sure it is worth it. You can create the same effect with `WindowMac.setTitle(null)` and `WindowMac.setFullSizeContentView(true)`.

What is the policy on what goes in `WindowMac` vs existing only in JNI methods?


mworzala:
> I am still able to drag the window near the top when using a full size content area.
This has to do with title visibility. If the title is invisible you cannot drag anywhere. If the title is visible, then you can drag anywhere at or above the title height. The subtitle is not included (eg you cannot drag over the subtitle)

tonsky:
Thank you for your quick PR! I merged it with minor change: I renamed constants in WindowMacTitlebarStyle to match macOS names.

I also added two ways to test this. First is `Cmd+T`, it switches titlebar of current window. Second is `Cmd+N`, each new window will be created with new titlebar style (there are 7 total, see `PanelScreens.java`).

These are different codepaths and we need them both to work. But I feel like the experience of changing window style on the fly is not that important. Most of the time you set the style before creating window (what `Cmd+N` tests).

I also realized that `setTitle(null)` is not such a great idea because title is used in e.g. Expose. I guess we’ll need another way of adding traffic light to titleless window. I am also happy to only allowing this when titlebar is hidden (makes little sense in all other cases).

Couple of other issues:

- I think we need to reset traffic lights position every time `setTitlebarStyle` is called.
- Right now `setTrafficLightPosition` specifies top left corner of traffic lights. I think we need a way to know its size, e.g. `IRect getTrafficLightBounds()` that will return both current position and current size. Size is needed to center it properly.
- `setFullSizeContentView(true)` does not work properly after `setTitlebarVisible(true)` first time, but works ok the next time around (`"Transparent"` mode in `PanelScreens`). It almost feels like `setTitlebarVisible` is async and takes a lot of time to complete, but I might be wrong.
- I added artificial ResizeEvent each time `setTitlebarStyle`/`setTitlebarVisible`/`setFullSizeContentView` are called
- Example also calls `Window#focus()` after each style change, because otherwise window loses focus
- `Window#focus()` does not work on a window with a hidden titlebar. I guess hiding titlebar makes window somehow special. We’ll need to find a way to make it a normal window again.
- If window starts with `window.setTitlebarVisible(false);` (before `window.setVisible(true);`), it stops receiving keyboard events. Clicking on it doesn’t work.

I have some experience trying to tame traffic light. I think what Electron does is the most reliable way. IIRC it puts default buttons into its own View and then positions that container in main window. Positioning individual buttons won’t work. Another crucial (and very obscure) detail is that container that contains those buttons must respond to some selectors, otherwise it will behave weird. See https://stackoverflow.com/a/30417372/142655

I would greatly appreciate help with any of that.

mworzala:
If you have experience I will let you tackle the traffic light positioning, this is my first time trying to modify them and I am not surprised I missed something.

Is there any reason we cannot treat `window.setTitlebarVisible(false)` as simply full size content view + hide title + hide traffic lights instead of actually removing `NSWindowStyleMaskTitled`? (which as you mentioned causes weird behavior and it removes corner rounding on the window)

This does create some potentially weird cases where you could do something like
```java
window.setTitlebarVisible(false);
window.setFullSizeContentView(false);
```
and then the titlebar would come back. Of course this could be solved by ignoring any calls to full size content view while the titlebar is gone.

>  because title is used in e.g. Expose

I am not sure what this means. I also do not have a better idea for getting rid of the title than `WindowMac#setTitleVisible(boolean)`, but it still has an exclusivity issue with invisible titlebar.

Both the `Window#focus()` and lack of keyboard input with invisible titlebar appear to be caused by [this](https://stackoverflow.com/a/30502481). Looks like a simple fix, however, it would also be fixed by my previous comment regarding not actually removing the titled style.

tonsky:
> If you have experience I will let you tackle the traffic light positioning, this is my first time trying to modify them and I am not surprised I missed something.

Got it https://github.com/HumbleUI/JWM/issues/183

> Is there any reason we cannot treat window.setTitlebarVisible(false) as simply full size content view + hide title + hide traffic lights instead of actually removing NSWindowStyleMaskTitled?

That’s what we probably should do, yes. Is there a NSWindow method to hide window title instead of setting it to empty string? I am worried about latter, because window _should_ keep its title, it should just be invisible. If yes, I am happy.

Just to clarify: we keep `window.setTitlebarVisible(false);` in Java API (because it’s cross-platform) but under-the-hood it will hide title, traffic light and make full sized content view, right?

mworzala:
> Is there a NSWindow method to hide window title instead of setting it to empty string

Its possible I'm misunderstanding, but there is the [`titleVisibility`](https://developer.apple.com/documentation/appkit/nswindow/1419635-titlevisibility) property which can be used to hide the titlebar (without touching the underlying string). I am not sure how it works internally.

> Just to clarify: we keep window.setTitlebarVisible(false); in Java API (because it’s cross-platform) but under-the-hood it will hide title, traffic light and make full sized content view, right?

Yep this is what I was thinking, I can work on this.

tonsky:
> titleVisibility

Sounds like just what we need!

Issue #183, opened 2022-01-14
macOS: custom traffic light position


tonsky:
I think what Electron does is the most reliable way. IIRC it puts default buttons into its own View and then positions that container in main window. Positioning individual buttons won’t work. Another crucial (and very obscure) detail is that container that contains those buttons must respond to some selectors, otherwise it will behave weird. See https://stackoverflow.com/a/30417372/142655

PR #184, opened 2022-01-26
More macOS titlebar changes

mworzala:
Makes the following changes:
 * Do not actually remove the title from the NSWindow, instead hide the title, traffic lights, and set full size content view
 * Add `WindowMac#setTitleVisible(boolean)`

Seeing an issue with setTitlebarVisible currently, and waiting on a way to hide traffic lights from #183

### DIFF
--- a/linux/java/WindowX11.java
+++ b/linux/java/WindowX11.java
@@ -72,7 +72,7 @@ public class WindowX11 extends Window {
     public Window setTitle(String title) {
         assert _onUIThread();
         try {
-            _nSetTitle(title == null ? new byte[0] : title.getBytes("UTF-8"));
+            _nSetTitle(title.getBytes("UTF-8"));
         } catch (UnsupportedEncodingException ignored) {}
         return this;
     }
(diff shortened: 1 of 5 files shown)


tonsky:
I will probably postpone #183 to way later since it’s not that high of the priority. If you can just hide traffic lights really quick without changing their position, just go ahead, I don’t want to be a blocker

mworzala:
I added a temporary way to toggle the traffic lights. I am unsure what is going on with the drawing rect. When setting full size content view the titlebar goes away, but the content is not shifted up to fill the gap until you resize the window manually. Then it seems to work from then on out.
eg. CMD-T and note the gap at the top, then resize and it is fixed. The issue appears on `main` as well, but it's less obvious (It happens in the `Transparent` setting).

tonsky:
I suggest we focus on creating new windows, not switching styles on existing one. I think it’s a more realistic use-case (you usually know what type of window you want, instead of deciding on the fly).

For that use-case this PR seems to work good, so let’s merge it.

PR #187, opened 2022-01-30
X11: fix back/forward mouse buttons (closes #145)

### LINKED ISSUE #145: X11: Back/forward mouse buttons are reported as PRIMARY

dzaima:
Use the correct (at least on my Linux) button numbers; also ignores 4 & 5, because those are mouse wheel up/down while JWM doesn't have those.


tonsky:
Interesting — will check on my Linux too and merge

tonsky:
Thank you! Merged

PR #196, opened 2022-02-03
X11: Copy windows for iteration

dzaima:
Otherwise the `EventFrame` dispatch may close the item the for loop currently holds an iterator of, ending up in undefined behavior and segfaults.

Alternative would be to keep a vector of windows to remove in `WindowManagerX11`, and removing them all inside `_processCallbacks` after they've been iterated through.


tonsky:
Thank you!

PR #199, opened 2022-02-08
X11: Blocking event loop

### LINKED ISSUE #129: Get rid of sleep(0.5ms) in X11 event loop
https://github.com/JetBrains/JWM/blob/c6762d5c61a8cb06dae5023ca43ddd43b87698cf/linux/cc/WindowManagerX11.cc#L195

Problem:

- Can’t use blocking event loop because no way to schedule runOnUIThread from another threads
- Non-blocking event loop eats lots of CPU

Ideas:

- don’t sleep at all
- provide user-defined lambda to call instead of sleeping

dzaima:
closes #129

(also makes the X error handler use stderr and flush for a much better development experience)

### DIFF
--- a/linux/cc/WindowManagerX11.cc
+++ b/linux/cc/WindowManagerX11.cc
@@ -5,6 +5,9 @@
 #include <impl/Library.hh>
 #include <impl/JNILocal.hh>
 #include "AppX11.hh"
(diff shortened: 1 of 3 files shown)


tonsky:
Awesome! Thank you!

PR #200, opened 2022-02-08
windows:support focus events

i10416:
JWM dispatches WindowFocusOn and WindowFocusOff events on Windows.
In addition, JWM checks if window exists before dispatching events to
avoid NPE.


tonsky:
Awesome! Thank you!

PR #204, opened 2022-02-08
support windows visibility events

i10416:
I temporarily named event as `EventWindowAppear/Disappear`, but I think there would be better candidates.

Which do you think is better for visibility event name?

- EventWindowShow/Hidden
- EventWindowAppear/Disappear
- EventWindowVisibilityOn/Off
- EventWindowVisible/Invisible
- other

JWM notifies when window's visibility changes. In Windows, visibility is based mainly on WS_VISIBLE property and
updated by ShowWindow function and ShowOwnedPopups function.

see
- https://github.com/MicrosoftDocs/win32/blob/docs/desktop-src/winmsg/window-features.md#window-visibility
- https://docs.microsoft.com/en-us/windows/win32/winmsg/wm-showwindow


tonsky:
How does it relate to minimize/maximize/restore (we already have those)? Minimize == hidden? Restore == visible? Or is there a subtle difference?

i10416:
> How does it relate to minimize/maximize/restore (we already have those)? Minimize == hidden? Restore == visible? Or is there a subtle difference?

It is not related to actual window visibility.  It reacts to ShowWindow function and ShowOwnedPopups function, but does not react to manually resizing(hiding) window🤔

PR #205, opened 2022-02-08
macOS App Progress Bar

### LINKED ISSUE #191: Progress bar API
See https://www.electronjs.org/docs/latest/tutorial/progress-bar

I see two options for an API here.
1. Electrons route, a `Window` method to set the current progress. The problem here is that Windows is the only platform capable of a progress bar per window.
2. An `App` method to set the progress for the entire app, with an overload to provide the window (or a secondary method on window to set it per window).

Either way it is confusing on some platforms.

mworzala:
Adds a macOS implementation of an app progress bar.

Resolves #191

### DIFF
--- a/examples/dashboard/java/Example.java
+++ b/examples/dashboard/java/Example.java
@@ -35,6 +35,8 @@ public class Example implements Consumer<Event> {
 
     public boolean paused = true;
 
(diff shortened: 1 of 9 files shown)


tonsky:
Thank you! That’s awesome

PR #207, opened 2022-02-09
X11: support input methods

### LINKED ISSUE #206: X11: EventTextInput doesn't respect XCompose
Entering a compose sequence (e.g. `<compose> 2 3`) won't result in `⅔` being inserted, but rather `2` and `3` separately.

dzaima:
Lets input methods consume key press events to build compose/dead key sequences. Additionally keeps the currently focused window as the input context focus, otherwise multiwindow applications would only allow IMs to work on the newest opened window.

Closes #206

### DIFF
--- a/linux/cc/WindowManagerX11.cc
+++ b/linux/cc/WindowManagerX11.cc
@@ -530,6 +530,7 @@ void WindowManagerX11::_processXEvent(XEvent& ev) {
         }
 
         case FocusIn: { // focused
+            XSetICFocus(myWindow->_ic);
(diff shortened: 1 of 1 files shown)


tonsky:
Fantastic, thank you! I feel like we've made so much progress in a last few weeks, thanks to your contributions, @i10416 and @mworzala. Thank you all folks!

dzaima:
I've just been working on making my programs work as well with JWM as they do with LWJGL (or better in some cases) :)

tonsky:
BTW if you have something ready/coming out, I’d be happy to put it in the README under “Projects using JWM”, just let me know!

PR #209, opened 2022-02-10
macOS: Menu bar issues

mworzala:
When launching the example app, you are unable to interact with the menu bar (including the "apple menu").

The problematic call is `[NSApp activateIgnoringOtherApps:YES]` in `WindowMac#setVisible`. When called before the app has finished launching this fails. A solution proposed in [this SO answer](https://stackoverflow.com/questions/62739862/why-doesnt-activateignoringotherapps-enable-the-menu-bar) is to add an app delegate which calls `activateIgnoringOtherApps` on `applicationDidFinishLaunching` which does solve the issue.

The basic app menu does work after this change.

### DIFF
--- a/macos/cc/App.mm
+++ b/macos/cc/App.mm
@@ -4,6 +4,7 @@
 #include "impl/Library.hh"
 #include "MainView.hh"
 #include "Util.hh"
+#include "ApplicationDelegate.hh"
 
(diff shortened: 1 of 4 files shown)


tonsky:
You are a savior! Thank you!

PR #213, opened 2022-02-13
Workaround `-XstartOnMainThread` requirement on macos

### LINKED ISSUE #210: macOS: Work around for -XstartOnMainThread
JavaFX appears to use [`performSelectorOnMainThread`](https://developer.apple.com/documentation/objectivec/nsobject/1414900-performselectoronmainthread) to start the app, perhaps it is worth doing something similar. Should also be easier to work with native image, as apparently the option [does not work](https://github.com/oracle/graal/issues/1861) (though I have not tested personally)

Eg 
```java
App.init(() -> {
    Window window = App.makeWindow();
    // ...

    App.start();
});
```

mworzala:
The following does not require the start parameter to be set, and will move to the main thread for init when not already on it. The caveat being that all user init must be done in the callback added to `App#init`. 

!! I was _not_ able to test the changes on the linux/windows side. 

As always comments are very welcome, I have not worked with JNI much so it's very possible I made mistakes.

Also, `init` does not seem like a particularly descriptive name for `App#init` anymore.

Closes #210

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,6 @@ Alpha. Expect API breakages.
 |-------------------|---------|-------|-----|
 | init              | ✅   | ✅   | ✅   |
 | makeWindow        | ✅   | [#121](https://github.com/HumbleUI/JWM/issues/121)   | ✅   |
-| start             | ✅   | ✅   | ✅   |
 | getScreens        | ✅   | ✅   | ✅   |
 | getPrimaryScreen  | ✅   | ✅   | ✅   |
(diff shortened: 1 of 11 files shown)


mworzala:
One more note: I have not tried building it with native-image (no arm build), but using `jpackage` works fine (not sure if it had issues prior).

tonsky:
Thanks, seems to work nice! It’s going to simplify Clojure REPL-based workflows so much!

PR #215, opened 2022-02-15
X11: fix clipboard

dzaima:
It's weird that it worked at all before (didn't work for pasting in gnome-terminal & lwjgl though, which made me investigate)


tonsky:
Thank you! Works for me in Terminal now too

Issue #186, opened 2022-01-30
X11: Alt key stays pressed after alt-tabbing away & back


dzaima:
One fix would be to clear `KeyX11.cc` state on the window being unfocused; and maybe keeping a set of pressed keys to all give an `EventKey` release on window unfocus. But maybe there's a better way to do this

PR #216, opened 2022-02-15
X11: clear modifiers on window unfocusing

dzaima:
A temporary-ish fix for #186; still leaves a dangling `KeyPressed` without a corresponding `KeyReleased` if you press a key, tab out, release key, tab back in. Fixing that would take some other much more complicated way.


tonsky:
Thanks! Since we are resetting internal state, I figured it will make sense to send EventKey up too. Merged in 755685858cc748c813b2a614401247052e7d6901

PR #220, opened 2022-03-21
X11: reset scroll distance after re-entering window

dzaima:
Does the same thing as other apps, resetting the scrolled distance every time the window is re-entered.

### DIFF
--- a/linux/cc/WindowManagerX11.cc
+++ b/linux/cc/WindowManagerX11.cc
@@ -314,6 +314,30 @@ void WindowManagerX11::_processXEvent(XEvent& ev) {
                         _xi2IterateDevices();
                         break;
 
+                    case XI_Enter: {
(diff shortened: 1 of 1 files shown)


tonsky:
Thanks! I was unable to repro this on Ubuntu, I guess it already accounts for that. First scroll is indeed ignored

dzaima:
Huh. May be possible to check for OS versions or something, but I don't know how would one go about doing that, or if it's even worth it to save that single scroll event. (didn't ubuntu switch to wayland though, meaning JWM runs on xwayland, which'd do different things than a global X server? Even then, there may still be issues with two open JWM windows within the same Java/X instance)

tonsky:
I don’t think it’s worth saving. I has tested on X11 Ubuntu, not on Wayland

Issue #51, opened 2021-07-19
macOS: Implement Clipboard API


tonsky:

lgorence:
@tonsky Is this something that you're working on? I see that you're assigned, but if you're not active on I'd like to take it on.

bumfo:
> @tonsky Is this something that you're working on? I see that you're assigned, but if you're not active on I'd like to take it on.

Are there any progress with this? I'm glad to help if there are some unfinished work.

tonsky:
No progress, I’m focusing on the https://github.com/HumbleUI/HumbleUI right now. Any help will be appreciated

lgorence:
> > @tonsky Is this something that you're working on? I see that you're assigned, but if you're not active on I'd like to take it on.
> 
> Are there any progress with this? I'm glad to help if there are some unfinished work.

Didn't make much progress on my end, been busy finishing up my last semester of university. Feel free to take it on though! 🙂
I would be glad to see this work done.

bumfo:
text/plain support has been implemented in #225. Not sure how to implement other formats yet.

PR #225, opened 2022-04-09
mac clipboard: text/plain support

bumfo:
A partial implementation of #51. Formats other than text/plain is not supported yet.

### DIFF
--- a/macos/cc/ClipboardMac.hh
+++ b/macos/cc/ClipboardMac.hh
@@ -1,0 +1,1 @@
+#pragma once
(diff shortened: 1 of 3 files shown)


lgorence (approved):
Thanks for picking up on this work! 🙂

tonsky:
Thanks! Merged as c32163c78536cb38c54c524392489ba0855727a5

bumfo:
> Thanks! Merged as [5215c56](https://github.com/HumbleUI/JWM/commit/5215c56841dcc111de964b1359fbc3ea7bc2804d)

Thanks for fixing. I would be more careful about JNILocal the next time. ;)

tonsky:
They are tricky—sometimes they matter, sometimes they don’t. The main purpose of them is to clean up local references in loops, but I prefer to use them everywhere just to be sure

PR #230, opened 2022-04-17
Check whether is closed before invoking native methods

### LINKED ISSUE #229: Overridden close closes before checking if already closed
The `close` method in `Managed` has a check for whether `_ptr` is zero, and throws an error if so. But if the method is overridden, e.g.
https://github.com/HumbleUI/JWM/blob/b3f2289ce4635ce53219cddd00809e7fdcc4008c/shared/java/LayerGL.java#L59-L64
(this kind of thing happens in most overridden `close` methods) then the native close method will be called before that, resulting in a segfault, which is a lot less clear on what has happened.

This is a pretty minor thing all things considered, as it only affects code that's already incorrect, but I think it's still worth fixing.

Maybe replace `assert _onUIThread();` with `assert _onUIThread() && !isClosed();`?

dzaima:
closes #229

### DIFF
--- a/linux/java/WindowX11.java
+++ b/linux/java/WindowX11.java
@@ -113,7 +113,7 @@ public class WindowX11 extends Window {
 
     @Override
     public void close() {
-        assert _onUIThread();
+        assert _onUIThread() && !isClosed();
         _nClose();
         super.close();
     }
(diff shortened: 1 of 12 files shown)


tonsky:
Thanks!

PR #231, opened 2022-05-27
Resizing mouse cursors

dzaima:
warning: completely untested on macOS or windows as I don't have access to those, but I tried to find some names for the icons for those OSes anyway.

Adds resizing cursors (plus more X11 definitions for the existing ones). Names taken from [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor).

### DIFF
--- a/linux/cc/WindowManagerX11.cc
+++ b/linux/cc/WindowManagerX11.cc
@@ -165,13 +165,25 @@ WindowManagerX11::WindowManagerX11():
             return XCreateFontCursor(display, alternativeId);
         };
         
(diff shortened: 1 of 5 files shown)


tonsky:
Thanks! Merged these four:

```
RESIZE_NS     = 8,
RESIZE_WE     = 9,
RESIZE_NESW   = 10,
RESIZE_NWSE   = 11,
```

The rest wasn’t working on any of my systems. Mac has singular N/S/W/E, but neither linux nor windows seem to have those. Also mac is missing NESW/NWSE right now. I guess the solution here would be to provide custom cursors (I think that’s what both browsers and AWT do), then we can cover all the use-cases

tonsky:
I tried on Ubuntu 20.04

PR #233, opened 2022-06-22
macOS: Implement window focus events

LuisThiamNye:
Implements the EventWindowFocusIn and EventWindowFocusOut events on macOS. 

It is worth noting that macOS has different types of focus: whether the application is active, whether a window accepts input (key window) and whether a window is the main window (has the active appearance). This implementation listens to 'key window' events, because this gives expectations consistent with the X11 and Windows implementations.


tonsky:
Dope, thank you!

PR #234, opened 2022-06-27
macOS: Fix cursor resetting to arrow

LuisThiamNye:
After using `setMouseCursor`, I noticed that when the cursor leaves the window and re-enters, it returns to the default arrow cursor.

The solutions to this are to use either `addCursorRect:cursor:` or `NSTrackingArea` to ensure the cursor is set back to the preferred image. This implementation builds on the existing tracking area, which I think is better than [an implementation using cursor rects](https://github.com/HumbleUI/JWM/compare/0.4.4...LuisThiamNye:JWM:luistn/cursors?expand=1).

Also, this patch includes a small modification to the dashboard example which allows you to click to the mouse cursor panel to keep the current cursor set until another element sets it back. As a result, the effect of this patch can be clearly seen.


tonsky:
Thank you!

PR #237, opened 2022-09-05
Point to supported Skija in readme

frenchy64:
I'm just learning about Skija+JWM, should this link point to the new repo?


tonsky:
Of course! Thanks

PR #238, opened 2022-09-13
linux: avoid crash on empty X11 resources

einars:
Running under XWayland, or just under my setup, XResourceManagerString returns null. This avoids the XrmGetStringDatabase crash.


tonsky:
Thanks! Pushed 0.4.8

PR #242, opened 2022-09-17
X11: Update mouse position in XI_Motion event (closes #241)

### LINKED ISSUE #241: X11: EventMouseScroll and EventMouseButton report incorrect cursor position
These events report incorrect, rarely updated `_x` and `_y` values.
events

I've been able to track down the source of this issue: unlike `MotionNotify` handler of `XI_Motion` doesnt update  `lastMousePosX` and `lastMousePosY` before dispatching `EventMouseMove`. Handler of `MotionNotify` does set these values but I find that it's very rarely invoked. This results in scroll and click events often reporting incorrect mouse position.

This can be fixed by replicating behaviour of `MotionNotify` handler and updating last mouse position before event dispatch.
fix.
events after fix

navaratine:
Implements the fix proposed in #241


tonsky:
Thanks!

Issue #243, opened 2022-09-17
X11: Scroll not reported while some other mouse button is held down


dzaima:
So far I've traced it down to [this](https://github.com/HumbleUI/JWM/blob/8814b93ae484e469097530f99cd35bfdadd5f891/linux/cc/WindowManagerX11.cc#L334-L337):

While no mouse buttons are held, that gets two events, one with `deviceEvent->deviceid` being `2` and one with it being `10`; the `10` fails the `if` and thus doesn't get processed.

Whereas when some mouse button is held down, only the `10==deviceEvent->deviceid` event happens, and so nothing gets processed.

Hard-coding `deviceEvent->deviceid` there to always be `2` does give scroll updates, proving that receiving scroll updates while a button is held is possible, but of course such hard-coding won't work for an actual solution.

jakeisnt:
Looks like this should be closed?

dzaima:
yep

PR #244, opened 2022-09-17
X11: Potential fix for #243

dzaima:
see #243

Seems to work for me (linux mint 21; don't have a touchpad to test to see how this may affect it), but I don't know if this is a good approach. `classInfo->sourceid` contains the id that is received both when some mouse button is and isn't held, so this switches to using it to look up the device instead of deviceid.

(also renames loop variable to avoid confusion as it's unrelated to the actual deviceid)


navaratine:
I've checked the behaviour with touchpad (OpenSuse) and patch seems to work just fine.
image

tonsky:
Doesn’t solve the issue for me on Ubuntu, but doesn’t seem to make anything worse either

PR #247, opened 2022-09-25
X11: Clear all scroll valuators on XI_Enter

dzaima:
My previous X11 PR (#244) introduced an issue where tabbing out & back into a window didn't clear the scroll valuator (just moving mouse out & in did work, which is what I had tested). I don't see a way to clear them precisely here, as neither the sourceid nor deviceid is the ID that stores the valuator, without introducing a map from device to source IDs, so this patch just clears all unconditionally.

Given that the IDs are the same between multiple JWM windows (at least on my Linux Mint 21 system), I don't think this should affect anything either.

### DIFF
--- a/linux/cc/WindowManagerX11.cc
+++ b/linux/cc/WindowManagerX11.cc
@@ -331,15 +331,10 @@ void WindowManagerX11::_processXEvent(XEvent& ev) {
                             break;
                         }
 
-                        auto itMyDevice = _xi2->deviceById.find(deviceEvent->sourceid);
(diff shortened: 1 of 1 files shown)


tonsky:
I won’t pretend I understand what’s going on, so hopefully it fixes something for you :) Thanks!

dzaima:
I won't pretend to understand much of what's going on either; X11 is weird, badly documented, and in general a pain to work with, but I want JWM to work well on it so I'm trying my best.

tonsky:
Thank you for that, it wouldn’t be as good without your help!

Issue #249, opened 2022-11-24
macOS: trackpad touch events


shaunlebron:
Looking at adding access to macOS trackpad touch events.  Some references to start:

* [Magic]: an example mac app that uses trackpad touch events for drawing
* [docs on macos touch events]( https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/HandlingTouchEvents/HandlingTouchEvents.html#//apple_ref/doc/uid/10000060i-CH13-SW1)
* edit: [docs on web touch events](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events)
* edit: [how SDL tracks touch events on macOS](https://github.com/libsdl-org/SDL/blob/a463aca0d1353be3fa45e09429b378cd28eb5a59/src/video/cocoa/SDL_cocoawindow.m#L1378-L1512)

[Magic]:https://apps.apple.com/us/app/magic-drawing-app/id1581223559?mt=12

tonsky:
I think web api should be studied for this too

shaunlebron:
The [Apple docs on touch events][doc] said to use `setAcceptsTouchEvents:YES`, but the compiler said it was deprecated for [allowedTouchTypes].  I had to find a [usage example] on GitHub.

The `acceptsTouchEvents` property also seemed necessary.  Does Apple have newer docs on this?

I confirmed that the touch events are being caught now (see 49a4483cda4619b45133482f95cee91e2620de32).  I will look next look at logging the touch event data to the window.

[usage example]:https://github.com/m1rr0r-t34m/SwipeTranslate/blob/ecf7e8190a79587f8a49dcb6906bbf86a7eed1d6/googleTranslateWidget/STDraggableView.m#L109-L115
[docs]:https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/HandlingTouchEvents/HandlingTouchEvents.html#//apple_ref/doc/uid/10000060i-CH13-SW1
[allowedTouchTypes]:https://developer.apple.com/documentation/appkit/nsview/2544839-allowedtouchtypes?language=objc

shaunlebron:
I added `EventTrackpadTouch*` events and the dashboard event panel is logging them successfully.  I will look at drawing the touch points in a new trackpad panel.

tonsky:
That would be cool!

shaunlebron:
I put a trackpad panel next to the mouse panel:
[...]

PR #254, opened 2022-12-09
macOS: cancel absent touches

shaunlebron:
Wraps up #249

In the dashboard, if you do a four-finger swipe up then return to the window, the four touch points never get a chance to unregister and just stick around.  This fixes it by removing all touch points not present in the active touch event.


tonsky:
Great, I think this is a great approach

tonsky:
Cut a new version 0.4.11, should arrive to Maven Central in ~30-ish minutes

PR #255, opened 2022-12-12
macOS: add EventTouchFrame

shaunlebron:
Hopefully last thing for #249.

This event allows us to do something at the end of a touch frame.  e.g. PanelTouch now uses it to request a redraw after it receives all the touch events for that frame.

CleanShot 2022-12-12 at 15 30 04@2x

### DIFF
--- a/examples/dashboard/java/PanelTouch.java
+++ b/examples/dashboard/java/PanelTouch.java
@@ -25,15 +25,13 @@ public class PanelTouch extends Panel {
             touchDevices.put(ee.getDeviceId(), ee.getId());
             devices.put(ee.getDeviceId(), new Point(ee.getDeviceWidth(), ee.getDeviceHeight()));
             touches.put(ee.getId(), new Point(ee.getFracX(), ee.getFracY()));
-            window.requestFrame();
         } else if (e instanceof EventTouchMove ee) {
             touches.put(ee.getId(), new Point(ee.getFracX(), ee.getFracY()));
-            window.requestFrame();
         } else if (e instanceof EventTouchCancel ee) {
             touches.remove(ee.getId());
-            window.requestFrame();
         } else if (e instanceof EventTouchEnd ee) {
             touches.remove(ee.getId());
+        } else if (e instanceof EventTouchFrame ee) {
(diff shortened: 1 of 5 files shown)


tonsky:
Do you mind if I challenge this design a little? Neither AppKit nor Web seems to have anything like it. Usually only one touch event happens at once (either you moved your finger or added/removed one), no?

If you are worried about scheduling frames, multiple requestFrames are collapsed if painting didn’t happen yet.

Can you think of a use-case for this?

shaunlebron:
Sure, I don’t really like it either.  The Web API sends all touch events for a given frame all at once.  I didn’t know how to send an array of touches in this way, so the current API just sends them individually.  That’s why I added a frame end event.  Maybe I will work on a separate PR that collates them into an array on the Java side.

tonsky:
You mean they group new touch with all the stationary touches from before? even if they didn’t change?

shaunlebron:
Not stationary touches— there can be multiple “moved” touches per frame.

tonsky:
Oh I see, [TouchEvent.changedTouches](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/changedTouches) is a list, and there’s also [TouchEvent.touches](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/touches)

I can help with array. I still think we shouldn’t group “by frame” and send them as soon as macOS reports them to us

tonsky:
Does this help? https://github.com/HumbleUI/JWM/blob/382e7626f25f15b797dc693548da33e43501f936/linux/cc/AppX11.cc#L118-L125

To be clear: as soon as touchesBeganWithEvent/touchesMovedWithEvent/etc is received on ObjC side, you’ll produce and report an event to Java side, but that event could have an array of multiple “changed touches” (because of your logic that computes the difference, or maybe because user was very fast/moved multiple fingers) and also an array of “all currently active touches”. At least this is how I see it, let me know if you want it differently

shaunlebron:
That sounds good to me!  Sorry if “frame” was confusing— I only meant that we should signal concurrent touch movements as macOS reports them.

Issue #130, opened 2021-08-24
Fullscreen


tonsky:
EvenWindowFullscreen
Window::requestFullscreen(boolean)

PR #259, opened 2022-12-24
Support fullscreen on X11

jakeisnt:
* [ ] EventWindowFullScreenEnter
* [ ] EventWindowFullScreenExit
* [x] Window::setFullScreen(boolean)
* [x] Window::isFullScreen()

Unhandled:
- Whether fullscreen should lock exclusive fullscreen (i.e. hiding background pop-ups). I assume that exclusive fullscreen is desirable, but left a hardcoded flag and documentation.
- Whether window decoration flags should be handled separately from full-screen (X11 can add and remove window decorations, like borders and the header, without changing the window size - is it safe to assume window decoration state doesn't change?)
- On XWayland, there is a minor amount of screen tearing visible. I was unable to track this down (and it may not be a problem on X11), but it's perfectly functional.


jakeisnt:
The two Events aren't explicitly supported by Wayland - still tracking them down. OK to merge without them.

jakeisnt:
Helps resolve #130

tonsky:
Works for me on Ubuntu 20.04. Thanks!

PR #260, opened 2023-01-01
Rudimentary Wayland Support

jakeisnt:
WIP
Goal: Set up Wayland compilation and get window creation working:
| start       
| makeWindow     
| getScreens      
| getPrimaryScreen  
| runOnUIThread   
| terminate     
Document changes and identify/specify minimum viable Wayland dependencies.

Cherry-pick Nix commits out before merge if undesirable. 

I think we want to depend on `wlroots` but there is reason not to - not every WM implementer should have to use the `wlroots` toolkit.


jakeisnt:
Was unsuccessful here, won't be able to follow through. Closing. Hoping that the partial progress has helped in some way.

tonsky:
Thanks for the attempt!

PR #268, opened 2023-06-22
Update README.md with macos build & install tip

Quezion:
Usually I'd bundle this into another PR but the MacOS implementation of `bringToFront` is ugly and might take a few weeks to land.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,21 @@ Run examples without building (use version from the table above):
 ./script/run.py --jwm-version <version>
 ```
 
+Generate & install a local .jar file:
+
+```
+./script/install.py
+```
+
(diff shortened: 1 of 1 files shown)


Quezion:
Little update on this stale PR:
I'll clean up the Mac implementation of `bringToFront` this week & get it PR'd, either reusing this one or opening a new one.

Quezion:
For some reason, using a local JWM version with publicly released `bringToFront` code (see merged #269) is failing to correctly steal focus in Mac. The code runs but simply fails to steal focus -- calls to `isFront` afterwards always returns `false`. Perhaps this is a permissions issue that the local build works around?

I'm not sure yet but I'll need to PR `bringToFront` once I figure out a fix. In the meantime, I'm going to close this PR & bundle it into #286 to reduce PR churn.

PR #269, opened 2023-06-26
Add Mac bringToFront with NSApp activateIgnoringOtherApps

Quezion:
Also implement IsFront with NSWindow isMainWindow

Mac equivalent to PR #266, where these two functions work to get "launcher functionality" out of my HumbleUI program. I'm also still using JNativeHook to read Alt+Space from the user's keyboard while the launcher program is in the background & then forcing it to the front via `BringToFront()`

`IsFront()` is similarly used to determine whether the launcher is _already_ in the foreground so that the launcher can self-minimize.


tonsky:
Thanks!

PR #271, opened 2023-09-17
Update Skija version and fix skija_native name for non-macOS

dzaima:
Updates Skija to 0.116.1, and makes it use the e.g. `skija-linux-x64` repo path instead of `skija-linux`; that was broken for the previous 0.109.2 too actually.


tonsky:
Thanks!

PR #272, opened 2023-09-17
Use XQueryPointer mouse position instead of XI_Motion fields

dzaima:
Without this, on my system, the `XI_Motion` mouse events were always lagging one event behind (not just one frame, but from whenever was the last time the mouse moved, so the wrong mouse pointer could be kept indefinitely).

i.e., in this gif, whenever I'm moving the mouse back & forth a single pixel, the `XI_Motion:` logs show the previous position (i.e. moving one pixel right results in it the X coordinate decreasing).

gif

The `XQueryPointer` values appear to be up-to-date always, however, so this patch switches to using them instead.

This effectively reduces the mouse→display latency by at least 1 frame or 16ms (which, though not the cause for me finding this, is definitely a good benefit!)

### DIFF
--- a/linux/cc/WindowManagerX11.cc
+++ b/linux/cc/WindowManagerX11.cc
@@ -410,15 +410,17 @@ void WindowManagerX11::_processXEvent(XEvent& ev) {
                             unsigned mask;
                             ::Window unused1;
                             int unused2;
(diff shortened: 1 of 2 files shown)


dzaima:
Added another commit of updating the mouse position on cursor re-entry into the window (and extracting the code for the mouse processing to a function).

tonsky:
Thanks!

PR #274, opened 2023-10-04
X11: add held modifier info on mouse events when window isn't focused

dzaima:
Before this, moving the mouse or scrolling in JWM window while it isn't focused would always give no held modifiers. With this patch, the modifiers are read from the appropriate field in addition to the key state logic. (mouse/key presses don't need changes, as they only happen when the window is already focused)


tonsky:
Thanks!

PR #278, opened 2023-12-03
Wayland support

TheDrawingCoder-Gamer:
Untested (couldn't get javac to work)
Highly expect something here to be extremely broken. Would not be surprised if there was an instant segfault.
bc it's literally a new platform this touches a LOT of files. Will work on merging


dzaima:
running within Weston:
```c
$ ./script/run.py
-- Java home found automatically at /usr/lib/jvm/java-1.11.0-openjdk-amd64. Set JAVA_HOME environment variable to override.
-- Configuring done
-- Generating done
-- Build files have been written to: /mnt/linux2/git/jwm-wayland/x11/build
ninja: no work to do.
-- Java home found automatically at /usr/lib/jvm/java-1.11.0-openjdk-amd64. Set JAVA_HOME environment variable to override.
-- Configuring done
-- Generating done
-- Build files have been written to: /mnt/linux2/git/jwm-wayland/wayland/build
ninja: no work to do.
Exception in thread "main" java.lang.UnsatisfiedLinkError: 'void io.github.humbleui.jwm.impl.Library._nAfterLoad()'
        at io.github.humbleui.jwm.impl.Library._nAfterLoad(Native Method)
        at io.github.humbleui.jwm.impl.Library.load(Library.java:70)
        at io.github.humbleui.jwm.App.start(App.java:26)
        at io.github.humbleui.jwm.examples.Example.main(Example.java:227)
Traceback (most recent call last):
  File "/mnt/linux2/git/jwm-wayland/./script/run.py", line 79, in <module>
    sys.exit(main())
  File "/mnt/linux2/git/jwm-wayland/./script/run.py", line 65, in main
    subprocess.check_call([
  File "/usr/lib/python3.10/subprocess.py", line 369, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['java', '--class-path', '/home/dzaima/.m2/repository/org/projectlombok/lombok/1.18.22/lombok-1.18.22.jar:/home/dzaima/.m2/repository/o
rg/jetbrains/annotations/20.1.0/annotations-20.1.0.jar:/home/dzaima/.m2/repository/io/github/humbleui/skija-shared/0.116.1/skija-shared-0.116.1.jar:/home/dzaima/.m2/repository
/io/github/humbleui/types/0.2.0/types-0.2.0.jar:target/classes:wayland/build:x11/build:/home/dzaima/.m2/repository/io/github/humbleui/skija-linux-x64/0.116.1/skija-linux-x64-0
.116.1.jar:examples/dashboard/target/classes', '-Djava.awt.headless=true', '-enableassertions', '-enablesystemassertions', '-Dfile.encoding=UTF-8', '-Xcheck:jni', 'io.github.h
umbleui.jwm.examples.Example']' returned non-zero exit status 1.
```

dzaima:
and here's a bunch of changes, ranging from fixes to hacks to debug utils (i.e. do not just add all of this), to get `./script/run.py` to not crash in Weston (doesn't render anything though):

<details>
<summary>that</summary>

```diff
diff --git a/examples/dashboard/java/Example.java b/examples/dashboard/java/Example.java
index f8a2ee2..fd3e29a 100644
--- a/examples/dashboard/java/Example.java
+++ b/examples/dashboard/java/Example.java
@@ -55,8 +55,9 @@ public class Example implements Consumer<Event> {

         var scale = window.getScreen().getScale();
         int count = App._windows.size() - 1;
-        Screen screen = App.getScreens()[(count / 5) % App.getScreens().length];
-        IRect bounds = screen.getWorkArea();
+        // Screen screen = App.getScreens()[(count / 5) % App.getScreens().length];
+        // IRect bounds = screen.getWorkArea();
+        IRect bounds = new IRect(0, 0, 100, 100);

         window.setTitle("JWM Window #" + count);
         if (window instanceof WindowMac windowMac) {
@@ -202,6 +203,7 @@ public class Example implements Consumer<Event> {
                 }
             }
         } else if (e instanceof EventFrame) {
+            System.out.println("frame");
             if (!paused)
                 window.requestFrame();
         } else if (e instanceof EventFrameSkija ee) {
diff --git a/examples/dashboard/java/PanelRendering.java b/examples/dashboard/java/PanelRendering.java
index 5ab882b..fde41f2 100644
--- a/examples/dashboard/java/PanelRendering.java
+++ b/examples/dashboard/java/PanelRendering.java
@@ -34,6 +34,8 @@ public class PanelRendering extends Panel {
             layers = new String[] { "LayerD3D12Skija", "LayerGLSkija", "SkijaLayerRaster" };
         else if (Platform.CURRENT == Platform.X11)
             layers = new String[] { "LayerGLSkija", "LayerRasterSkija" };
+        else if (Platform.CURRENT == Platform.WAYLAND)
+            layers = new String[] { "LayerRasterSkija", "LayerGLSkija" };

         for (var layerName: layers)
             layersStatus.put(layerName, UNKNOWN);
diff --git a/script/build.py b/script/build.py
index 9954d82..589461f 100755
--- a/script/build.py
+++ b/script/build.py
@@ -4,7 +4,7 @@ import argparse, build_utils, common, glob, os, platform, subprocess, sys
 def build_native_system(system):
   os.chdir(common.basedir + "/" + system)
   subprocess.check_call(["cmake",
-    "-DCMAKE_BUILD_TYPE=Release",
+    "-DCMAKE_BUILD_TYPE=Debug",
     "-B", "build",
     "-G", "Ninja",
     "-DJWM_ARCH=" + build_utils.arch,
diff --git a/script/run.py b/script/run.py
index 98f3c0e..e90cbe7 100755
--- a/script/run.py
+++ b/script/run.py
@@ -10,9 +10,10 @@ def main():
   parser.add_argument('--skija-shared-jar', default=None)
   parser.add_argument('--skija-platform-jar', default=None)
   parser.add_argument('--types-dir', default=None)
+  parser.add_argument('--just-run', action='store_true')
   args = parser.parse_args()

-  if not args.jwm_version:
+  if not args.jwm_version and not args.just_run:
     build.main()

   if args.skija_dir:
diff --git a/shared/java/App.java b/shared/java/App.java
index 268ddc6..7a1c5a7 100644
--- a/shared/java/App.java
+++ b/shared/java/App.java
@@ -52,6 +52,8 @@ public class App {
             window = new WindowMac();
         else if (Platform.CURRENT == Platform.X11)
             window = new WindowX11();
+        else if (Platform.CURRENT == Platform.WAYLAND)
+            window = new WindowWayland();
         else
             throw new RuntimeException("Unsupported platform: " + Platform.CURRENT);
         _windows.add(window);
diff --git a/shared/java/impl/Library.java b/shared/java/impl/Library.java
index 43fd5cb..29a5da5 100644
--- a/shared/java/impl/Library.java
+++ b/shared/java/impl/Library.java
@@ -46,10 +46,10 @@ public class Library {
         } else if (Platform.CURRENT == Platform.X11) {
             File library = _extract("/", "libjwm_x64.so", tempDir);
             System.load(library.getAbsolutePath());
-        } /*else if (Platform.CURRENT == Platform.WAYLAND) {
+        } else if (Platform.CURRENT == Platform.WAYLAND) {
             File library = _extract("/", "libjwm_x64_wayland.so", tempDir);
             System.load(library.getAbsolutePath());
-        }*/
+        }

         if (tempDir.exists() && version == null) {
             Runtime.getRuntime().addShutdownHook(new Thread(() -> {
diff --git a/wayland/CMakeLists.txt b/wayland/CMakeLists.txt
index 95ca0ad..f13c700 100644
--- a/wayland/CMakeLists.txt
+++ b/wayland/CMakeLists.txt
@@ -15,6 +15,8 @@ if(NOT JWM_ARCH)
     endif()
 endif()

+find_package(OpenGL REQUIRED)
+
 file(GLOB SOURCES_CXX ${CMAKE_CURRENT_LIST_DIR}/../shared/cc/*.cc
     ${CMAKE_CURRENT_LIST_DIR}/../linux/cc/*.cc
     ${CMAKE_CURRENT_LIST_DIR}/cc/*.cc )
@@ -44,3 +46,4 @@ set_target_properties(jwm PROPERTIES OUTPUT_NAME "jwm_${JWM_ARCH}_wayland")
 target_link_libraries(jwm PRIVATE ${WAYLAND_CLIENT_LIB} ${DECOR_LIB}
     ${WAYLAND_CURSOR} ${XKBCOMMON})
 target_link_libraries(jwm PRIVATE ${EGL} ${WAYLAND_EGL})
+target_link_libraries(jwm PRIVATE OpenGL::GL)
diff --git a/wayland/cc/LayerRasterWayland.cc b/wayland/cc/LayerRasterWayland.cc
index 0462940..2177fdf 100644
--- a/wayland/cc/LayerRasterWayland.cc
+++ b/wayland/cc/LayerRasterWayland.cc
@@ -36,7 +36,6 @@ namespace jwm {
             _pool->grow(bufSize);
             // LSBFirst means Little endian : )
             auto buf = _pool->createBuffer(0, width, height, width * sizeof(uint32_t), WL_SHM_FORMAT_ABGR8888);
-
             _buffer = buf.first;
             _imageData = buf.second;
         }
diff --git a/wayland/cc/WindowManagerWayland.cc b/wayland/cc/WindowManagerWayland.cc
index 2c62963..ee92779 100644
--- a/wayland/cc/WindowManagerWayland.cc
+++ b/wayland/cc/WindowManagerWayland.cc
@@ -75,7 +75,7 @@ WindowManagerWayland::WindowManagerWayland():
               .leave = WindowManagerWayland::pointerHandleLeave,
               .motion = WindowManagerWayland::pointerHandleMotion,
               .button = WindowManagerWayland::pointerHandleButton,
-              .axis = WindowManagerWayland::pointerHandleAxis
+              // .axis = WindowManagerWayland::pointerHandleAxis
             };
             wl_pointer_add_listener(pointer, &pointerListener, this);
         }
@@ -112,7 +112,7 @@ void WindowManagerWayland::runLoop() {
         }

         if (myPoll.revents & POLLIN) {
-            while (read(pipes[0], buf, sizeof(buf) == sizeof(buf))) {}
+            while (read(pipes[0], buf, sizeof(buf)) == sizeof(buf)) {}
         }
         notifyBool.store(false);
     }
@@ -166,7 +166,7 @@ void WindowManagerWayland::registryHandleGlobal(void* data, wl_registry *registr
                 &wl_shm_interface, 1);
     } else if (strcmp(interface, "xdg_wm_base") == 0) {
         self->xdgShell = (xdg_wm_base*)wl_registry_bind(registry, name,
-                &xdg_wm_base_interface, 2);
+                &xdg_wm_base_interface, 1);
     } else if (strcmp(interface, "wl_data_device_manager") == 0) {
         self->deviceManager = (wl_data_device_manager*)wl_registry_bind(registry, name,
                 &wl_data_device_manager_interface, 1);
diff --git a/wayland/cc/xdg-shell.cc b/wayland/cc/xdg-shell.cc
index 03826cd..4001974 100644
--- a/wayland/cc/xdg-shell.cc
+++ b/wayland/cc/xdg-shell.cc
@@ -28,16 +28,19 @@
  * DEALINGS IN THE SOFTWARE.
  */

+
 #include <stdlib.h>
 #include <stdint.h>
 #include "wayland-util.h"

+extern "C" {
+
 #ifndef __has_attribute
 # define __has_attribute(x) 0  /* Compatibility with non-clang compilers. */
 #endif

 #if (__has_attribute(visibility) || defined(__GNUC__) && __GNUC__ >= 4)
-#define WL_PRIVATE __attribute__ ((visibility("hidden")))
+#define WL_PRIVATE __attribute__ ((visibility("hidden"))) extern "C"
 #else
 #define WL_PRIVATE
 #endif
@@ -181,3 +184,5 @@ WL_PRIVATE const struct wl_interface xdg_popup_interface = {
 	3, xdg_popup_events,
 };

+
+}
diff --git a/wayland/java/WindowWayland.java b/wayland/java/WindowWayland.java
index 056f7f4..5c09dd5 100644
--- a/wayland/java/WindowWayland.java
+++ b/wayland/java/WindowWayland.java
@@ -32,14 +32,14 @@ public class WindowWayland extends Window {
     public IRect getWindowRect() {
         assert _onUIThread() : "Should be run on UI thread";
         // very very bad!
-        return IRect.makeXYWH(0, 0, 0, 0);
+        return IRect.makeXYWH(0, 0, 100, 100);
     }

     @Override
     public IRect getContentRect() {
         assert _onUIThread() : "Should be run on UI thread";
         // stop!
-        return IRect.makeXYWH(0, 0, 0, 0);
+        return IRect.makeXYWH(0, 0, 100, 100);
     }

     @Override
```

</details>

TheDrawingCoder-Gamer:
I have a question about what format the buffer is supposed to be in. I assumed ABGR8888 because of LSBFirst, but is it RGBA8888?

TheDrawingCoder-Gamer:
updates ( i guess):

window does now open, but no rendering
skija won't actually bind to LayerGL for some reason??? it can't find the egl context : /. idk what to do there
LayerRaster currently memory leaks every resize because skija is holding on to a pointer. I get crashes when uncommenting out `munmap` in `ShmPool::close`. 

I've made a surprising amount of progress which is nice. I want to get rendering working before I start hooking up the keyboard. Any help would be appreciated bc I don't fully understand C++ (multiple days were taken up because I didn't realize things were freed at the end of scope).

TheDrawingCoder-Gamer:
OH, you mean in skija. Yeah that is likely an upstream issue.

dzaima:
```java
#0  0x00007f22c3e5c9fe in glXGetCurrentContext () from /lib/x86_64-linux-gnu/libGLX.so.0
#1  0x00007f228af92132 in GrGLMakeGLXInterface() () from /tmp/skija_0.116.1_x64/libskija.so
#2  0x00007f228ae93d72 in GrGLMakeNativeInterface() () from /tmp/skija_0.116.1_x64/libskija.so
#3  0x00007f228aafe498 in GrGLGpu::Make(sk_sp<GrGLInterface const>, GrContextOptions const&, GrDirectContext*) () from /tmp/skija_0.116.1_x64/libskija.so
#4  0x00007f228a9e9a45 in GrDirectContext::MakeGL(sk_sp<GrGLInterface const>, GrContextOptions const&) () from /tmp/skija_0.116.1_x64/libskija.so
#5  0x00007f228a9e9ed7 in GrDirectContext::MakeGL() () from /tmp/skija_0.116.1_x64/libskija.so
#6  0x00007f228a4c2920 in Java_io_github_humbleui_skija_DirectContext__1nMakeGL () from /tmp/skija_0.116.1_x64/libskija.so
#7  0x00007f22b4356b11 in ?? ()
#8  0x00007f22cad16e10 in ?? ()
#9  0x0000000000000000 in ?? ()
```
the full gdb stacktrace of the fail fwiw

dzaima:
perhaps a `skia_use_egl=true` in like [here](https://github.com/HumbleUI/SkiaBuild/blob/bea683fb14a4bcb359d6019d6c3fad6c687897ee/script/build.py#L48) or something?

dzaima:
Got a skia build with that to load (ended up `ln -s`ing SkiaBuild's output unzipped to Skija's `platform/Skia-m116-f44dbc40d8-linux-Release-x64/` and doing a clean rebuild); gives this now:
```java
Exception in thread "main" java.lang.UnsatisfiedLinkError: /tmp/skija_369211854708109/libskija.so: /tmp/skija_369211854708109/libskija.so: undefined symbol: eglQueryString
	at java.base/jdk.internal.loader.NativeLibraries.load(Native Method)
	at java.base/jdk.internal.loader.NativeLibraries$NativeLibraryImpl.open(NativeLibraries.java:331)
	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:197)
	at java.base/jdk.internal.loader.NativeLibraries.loadLibrary(NativeLibraries.java:139)
	at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2404)
	at java.base/java.lang.Runtime.load0(Runtime.java:785)
	at java.base/java.lang.System.load(System.java:2011)
	at io.github.humbleui.skija.impl.Library._loadFromFile(Library.java:147)
	at io.github.humbleui.skija.impl.Library.load(Library.java:122)
	at io.github.humbleui.skija.impl.Library.staticLoad(Library.java:20)
	at io.github.humbleui.skija.Font.<clinit>(Font.java:9)
	at io.github.humbleui.jwm.examples.Example.<clinit>(Example.java:18)
```
should be easy enough to fix once I find where

dzaima:
([here's](https://github.com/HumbleUI/JWM/files/13590706/a.zip) a Skija linux x64 jar file of that; usable for JWM's example thing with `./script/run.py --skija-platform-jar /path/to/that/target/skija-linux-x64-0.0.0-SNAPSHOT.jar`)

dzaima:
Seems with this EGL build JWM & Skija still work on X11 with no extra changes? So that's nice (assuming that's correct and not me having forgotten to do something)

TheDrawingCoder-Gamer:
You did test X11, correct? this doesn't break X11 if I change skija?

dzaima:
yeah, said that in my last comment; running the exact same command works both in X11 and Weston, both using `LayerGLSkija` (with X11 properly rendering things as usual, and Weston quietly doing nothing). There could be some implications I'm not aware of, but ¯\\\_(ツ)\_/¯

TheDrawingCoder-Gamer:
While the window is infact showing, it's incapable of actually rendering anything. It can't actually make the DirectContext (it errors out, because `GrDirectContext::MakeGL().release()` returns `nullptr` only on wayland : /)

dzaima:
Huh? For me, `GrDirectContext::MakeGL().release()` gave `nullptr` before the `SkiaBuild` change, but the entire purpose of the Skia change was to make `GrDirectContext::MakeGL()` work, and it does appear to for me; i.e. adjusting Skija's `DirectContext::makeGL` to
```java
    public static DirectContext makeGL() {
        Stats.onNativeCall();
        System.out.println("pre-_nMakeGL in Java");
        long ptr = _nMakeGL();
        System.out.println("post-_nMakeGL in Java: got "+ptr);
        new Exception().printStackTrace();
        return new DirectContext(ptr);
    }
```
results in logging of
```
pre-_nMakeGL in Java
post-_nMakeGL in Java: got 140025466978336
java.lang.Exception
  at io.github.humbleui.skija.DirectContext.makeGL(DirectContext.java:15)
  at io.github.humbleui.jwm.skija.LayerGLSkija.frame(LayerGLSkija.java:27)
  at io.github.humbleui.jwm.Window.accept(Window.java:412)
  at io.github.humbleui.jwm.Window.accept(Window.java:427)
  at io.github.humbleui.jwm.Window.accept(Window.java:425)
  at io.github.humbleui.jwm.Window.setLayer(Window.java:66)
  at io.github.humbleui.jwm.examples.PanelRendering.changeLayer(PanelRendering.java:57)
  at io.github.humbleui.jwm.examples.PanelRendering.<init>(PanelRendering.java:44)
  at io.github.humbleui.jwm.examples.Example.<init>(Example.java:51)
  at io.github.humbleui.jwm.examples.Example.lambda$main$0(Example.java:229)
  at io.github.humbleui.jwm.App.lambda$start$1(App.java:35)
  at io.github.humbleui.jwm.App._nStart(Native Method)
  at io.github.humbleui.jwm.App.start(App.java:28)
  at io.github.humbleui.jwm.examples.Example.main(Example.java:228)
pre-_nMakeGL in Java
post-_nMakeGL in Java: got 140025466978336
java.lang.Exception
  at io.github.humbleui.skija.DirectContext.makeGL(DirectContext.java:15)
  at io.github.humbleui.jwm.skija.LayerGLSkija.frame(LayerGLSkija.java:27)
  at io.github.humbleui.jwm.Window.accept(Window.java:412)
  at io.github.humbleui.jwm.Window.accept(Window.java:427)
  at io.github.humbleui.jwm.Window.accept(Window.java:425)
  at io.github.humbleui.jwm.Window.setVisible(Window.java:280)
  at io.github.humbleui.jwm.WindowWayland.setVisible(WindowWayland.java:92)
  at io.github.humbleui.jwm.examples.Example.<init>(Example.java:88)
  at io.github.humbleui.jwm.examples.Example.lambda$main$0(Example.java:229)
  at io.github.humbleui.jwm.App.lambda$start$1(App.java:35)
  at io.github.humbleui.jwm.App._nStart(Native Method)
  at io.github.humbleui.jwm.App.start(App.java:28)
  at io.github.humbleui.jwm.examples.Example.main(Example.java:228)
```

TheDrawingCoder-Gamer:
hm, I must have messed up my local build. Did you do anything other than just add `skia_use_egl`?

TheDrawingCoder-Gamer:
Something must be different in our environments : / Adding that same code it still returns null under wayland, but works fine under x11.

dzaima:
[this](https://dzaima.github.io/paste/#0jZLfboIwFMbveYozdoGmln9DHSYk3u4ZjCGlHLAZVNKWiW8/DJItKNl60/b0d772O6e5KAqgtBQGmKe5Eo3xslZUudtcIZtGLCFz7CDz3wN/G7luGK/fwoxB4Pvb9dqilD6qWISQJ0r7PdAoXgUxkH4KfegDORZQMyEXy50Ft8FUqYEkcBi2AI7@FCxtNab6qg3WaaEQzbXBMDGqRWc1gq93FCXLKkzLph0BMlXCspokO9gZxVJeVKzUKefJwaaFMkbYxxE6WsMsCljUjJ@EREgScJiqN5GzBCbzn4OXBPi5rs/SlcyIL0zvB4vlaPS3VQrjM/qbbcYUP20iWgnZdrSULS05p7HdP@QX2HVPSUIGkvxDMggn5LzmgAI8K9etVh9eq5X3kOsJyas2x76KY/JxWGCl8c9KzPmecTnr6cHB0Eus@l46l/6Hny/auTVz@GK7bw#y0nMS9dQT8lMS1PXUXdW17QGAA) is my SkiaBuild diff (the compiler diffs are just because I don't have gcc-9 installed; shouldn't affect anything), and [this](https://dzaima.github.io/paste/#0tVVrb9pIFP3uX3HldiW7xuYRmleTKBScR@sAAiLtVitZgz2GgWFszYxpqir/fa8NKJAlSSvtWsJj3zlz7p2555iYJQm47oRpINWME52kclFt35E5DZjSytMPGsYvzRhMxPQBDscHHxsJ9bzGAUniIwr1Wu2w2TRc132Z1XAc5xXmy0twj5uVJjjlHV@piFli2YYBHFFWq9/3ux1o37W@@uFdr3Mf@GG/NbqB9z9XsWHvftD2w87t4LEaLZDcNoCJiOcxta6w8OGcEdtwNZETqkPOxBxvY0kko8pSczYjoOap1oziqCaglvibZ0SSiSTZtAhOSUYlPuSCRWlc4CRVaS4jqvCZEazlauD7o7/6fhjcfh60Brf@8BGjN63B1ef7b9@2orbh/K@l@NfB75UT75VGFFU7TNJIt1Oh6YP2omi7i3um1yKpN2mTRjXPOzk@rH08arwgkn0Ez5SyD1LIpVE5AadRadRQLQa8W/cazmaCedOLrYh5LXcJpqZhOJvpMzxCpeNyCU5SKcBsm/Cle@v/2e8NRjDjqZgU7@1WEMAXsiQhS0M8p2k@Dqf5YsxpzsKybeFOnjCsi0Ll14EBAFbBKJYfUNjLCswiTpRaDzb8NFyEgKQ6x/ySMiSQGb6GEVH6rCzhwnq2j9PTFbtle5JyShS1bPuT4RRMmUSKxDILa8B18LcwNzPqh4oI55Zb30RWG@Rw/t8m7hSJ4Y8Mc1fAWqYs/mDzDYwlYPHz81qx9TKwr7Antne/fEHvBro9GLY68OuLto4HIEl4rqYWaiLN9b4wlXITflwN677xTwZGCuUlMU1g@DW880etYNdaaF1J4@oMdfRM10UIhf86YG2vY1o/OK6hvcaExrWj2q693qBYGewNUGGxk8oROCeVer34IGf5mLMIVsLdWVD6RsQKBjRpC40tLU/sspvqbs45XBY4SSJtme4FCPrdtFeINaXSROOwy7lYS2zDVlxDBCovFV3EL2m7UAt2wt1Mr9uACXa5rFBsBPvUT2T7oTRdeNhlr9QZF5aJ0nc3aPz7KO1ubi8qvZJpiW55Yn2LNFX6X6ynMEk1mA5yba8vivcfIppplgo0WEmC@47mIzxBupPrxf2uOGGl0FeboadMmfY/#y0nMS9dQT8lMS1PXUXdW17QGAA) in Skija (all that matters is the `platform/CMakeLists.txt`; ignore the `syscall(-1)`, I abuse that as a "checkpoint" for debugging); after doing the SkiaBuild `script/build.py` and `script/archive.py`, I unzipped the `Skia-m116-f44dbc40d8-linux-Release-x64.zip` to a folder in Skija under `platform/Skia-m116-f44dbc40d8-linux-Release-x64`, removing what was there before. Then within Skija I ran `./script/clean.py` and then `./script/build.py && ./script/package_platform.py` and then for JWM's `--skija-platform-jar` used the output in Skija's `target/skija-linux-x64-0.0.0-SNAPSHOT.jar`.

In Weston I do have raster rendering the window.

dzaima:
here's a random stacktrace of a random segfault from making PanelRendering switch from raster to GL on a mouse button press (doesn't error always, sometimes it does manage to switch from raster to rendering nothing on GL; this is before the libdecor change):
```c
#0  0x00007f583a3efe43 in jwm::Window::dispatch(_jobject*) () from /mnt/linux2/git/jwm-wayland/target/classes/libjwm_x64_wayland.so
#1  0x00007f583a3f27c6 in jwm::WindowManagerWayland::mouseUpdate(jwm::WindowWayland*, unsigned int, unsigned int, unsigned int) () from /mnt/linux2/git/jwm-wayland/target/classes/libjwm_x64_wayland.so
#2  0x00007f583a327e2e in ?? () from /lib/x86_64-linux-gnu/libffi.so.8
#3  0x00007f583a324493 in ?? () from /lib/x86_64-linux-gnu/libffi.so.8
#4  0x00007f583a632ad0 in ?? () from /lib/x86_64-linux-gnu/libwayland-client.so.0
#5  0x00007f583a633243 in ?? () from /lib/x86_64-linux-gnu/libwayland-client.so.0
#6  0x00007f583a63343c in wl_display_dispatch_queue_pending () from /lib/x86_64-linux-gnu/libwayland-client.so.0
#7  0x00007f583a3f21dd in jwm::WindowManagerWayland::runLoop() () from /mnt/linux2/git/jwm-wayland/target/classes/libjwm_x64_wayland.so
```
definitely a worry for later, but worth noting regardless (error is from `myWindow` being null, presumably `pointerHandleMotion` of the old surface running after we've switched to the new one?)

will look at what the libdecor change looks like

dzaima:
ok with that last commit GL crashes with my "OH NO SAD", and raster doesn't render anything

TheDrawingCoder-Gamer:
It breaking may have been caused by me switching from `glext` api to `glplatform`. Unsure why that breaks stuff

dzaima:
with this, GL not crashing is restored:
```diff
diff --git a/wayland/cc/LayerGLWayland.cc b/wayland/cc/LayerGLWayland.cc
index 27f6c93..abc8aa9 100644
--- a/wayland/cc/LayerGLWayland.cc
+++ b/wayland/cc/LayerGLWayland.cc
@@ -73,6 +73,7 @@ namespace jwm {
                     throw std::runtime_error("Couldn't make context");
                 }
             }
+            makeCurrentForced();


         }
```

dzaima:
:o while undoing random parts of your changes, I now have a GL rendering the window in weston, and with proper refreshing too unlike raster! (those undoed changes include undoing all of libdecor though)

TheDrawingCoder-Gamer:
For me the _whole compositor_ must lose focus and be refocused for window to be visible. For SwayWM i am doing this by switching ttys, so something god awful is going on

dzaima:
Adding a `libdecor_dispatch(_windowManager.decorCtx, -1);` right after `libdecor_decorate` does something (and results in some segfaulting too; looking into that)

TheDrawingCoder-Gamer:
Try adding `libdecor_frame_map(_frame)` at the end of `show`, I'm not segfaulting with that at the end

TheDrawingCoder-Gamer:
I don't think I actually commited the change where I commented out the pointer listener, but you seem to be getting _something_, which is confusing

dzaima:
I'm getting a decorated window in Weston, but it only appears after changing the weston window size. And it's of course still segfaulting. I have a guess that there might be some memory corruption, am investigating

dzaima:
printing `fWindow` in `jwm::Window::dispatch` does give a different value on the crashing iteration. Honestly I have no clue what's happening (rr is breaking on me repeatedly), but potentially `libdecor_decorate` is writing over it for some reason..?

TheDrawingCoder-Gamer:
Is it possible it's losing all references for some reason?

dzaima:
the window object? maybe, though the `~Window` isn't called (libdecor specifically messing with it seems unlikely now; `xor`ing the `this` argument of the data pointers of `libdecor_decorate` and `wl_surface_add_listener` leads it to still failing)

TheDrawingCoder-Gamer:
????? the WHOLE this? wouldn't that extend to `WindowWayland` as well?

dzaima:
(ignore me there, I did a stupid)

dzaima:
libdecor does in fact not touch the user data pointer in any way other than passing it where it should

TheDrawingCoder-Gamer:
It adds listeners though, which also sets userdata.

dzaima:
ok the issue might be quite simple - the `window` derived in `pointerHandleMotion` is invalid (added a field of `int sanityCheck = 31415926;` in `WindowWayland`, and added `if (window->sanityCheck != 31415926) { printf("NOT WINDOW\n"); __builtin_trap(); }` in `pointerHandleMotion`)

TheDrawingCoder-Gamer:
I'll revert to finding in native : /. Bad optimization on my part

dzaima:
could it be that like the surface that `pointerHandleEnter` gets is the libdecor-internal-window surface, while the one set up in `WindowWayland::show` is the outer one that includes decoration, or something like that? or vice versa? (I have no clue)

TheDrawingCoder-Gamer:
If so I would assume vice versa, as I'm pretty sure that libdecor instances an outer surface to draw its frame.

TheDrawingCoder-Gamer:
The issue is that, if it IS a different surface, then changing back to the _nativeWindowToMy won't work bc it's a different pointer.

TheDrawingCoder-Gamer:
New commit fixes that. However I still have incorrect rendering

TheDrawingCoder-Gamer:
!!!
image
AND IT WORKS WITH HIDPI!

TheDrawingCoder-Gamer:
@dzaima, i'm curious on what debugging tools you are using. I am not really using debugging tools, which is why I'm struggling with making this stuff work.

TheDrawingCoder-Gamer:
todo tommorow:
* try harder with raster (it has many weird bugs)
* figure out why multiple windows doesn't work
* setting of clipboard

tbh i have no idea how i'm going to figure out number 2, bc it's kind of odd, and idk how to really inspect the app.
number 1 is also going to be a pain - switching layers isn't working most of the time, and raster only redraws whenever the compositor is like OH CRAP THIS IS SO MANY BUFFERS or an event occurs.

dzaima:
for whatever reason, raster works a lot better if `libdecor_dispatch` is called twice after `libdecor_frame_map`, not once. Feels like something that'd happen if some events are sent in the wrong order otherwise?

dzaima:
oh, and btw, changes I need to have it build (some that have been lingering around, some new:)

<details>
<summary>diff</summary>

```diff
diff --git a/script/package.py b/script/package.py
index 5bd1fa1..adde315 100755
--- a/script/package.py
+++ b/script/package.py
@@ -22,7 +22,8 @@ def main() -> Tuple[str, str, str]:
   jar = build_utils.jar(f"target/jwm-{common.version}.jar", ("target/classes", "."), ("target/maven", "META-INF"))
 
   build_utils.makedirs("target/src/io/github/humbleui/jwm")
-  shutil.copytree("linux/java", "target/src/io/github/humbleui/jwm", dirs_exist_ok=True)
+  shutil.copytree("x11/java", "target/src/io/github/humbleui/jwm", dirs_exist_ok=True)
+  shutil.copytree("wayland/java", "target/src/io/github/humbleui/jwm", dirs_exist_ok=True)
   shutil.copytree("macos/java", "target/src/io/github/humbleui/jwm", dirs_exist_ok=True)
   shutil.copytree("shared/java", "target/src/io/github/humbleui/jwm", dirs_exist_ok=True)
   shutil.copytree("windows/java", "target/src/io/github/humbleui/jwm", dirs_exist_ok=True)
diff --git a/wayland/cc/WindowManagerWayland.cc b/wayland/cc/WindowManagerWayland.cc
index f3ef205..193fadb 100644
--- a/wayland/cc/WindowManagerWayland.cc
+++ b/wayland/cc/WindowManagerWayland.cc
@@ -403,7 +403,7 @@ void WindowManagerWayland::keyboardKey(void* data, wl_keyboard* keyboard, uint32
     // TODO: while unlikely, it could be possible that you can be entering text even if the 
     // pointer hasn't entered
     if (self->keyboardFocus && jwmKey != jwm::Key::UNDEFINED) {
-        jwm::KeyLocation location;
+        jwm::KeyLocation location = jwm::KeyLocation::DEFAULT;
         JNILocal<jobject> keyEvent(
             app.getJniEnv(),
             classes::EventKey::make(
diff --git a/wayland/cc/WindowWayland.cc b/wayland/cc/WindowWayland.cc
index 083e0bc..5f643f2 100644
--- a/wayland/cc/WindowWayland.cc
+++ b/wayland/cc/WindowWayland.cc
@@ -14,8 +14,10 @@ using namespace jwm;
 wl_surface_listener WindowWayland::_surfaceListener = {
     .enter = WindowWayland::surfaceEnter,
     .leave = WindowWayland::surfaceLeave,
+    #ifdef HAVE_WAYLAND_1_22
     .preferred_buffer_scale = WindowWayland::surfacePreferredBufferScale,
     .preferred_buffer_transform = WindowWayland::surfacePreferredBufferTransform
+    #endif
 };
         
         
@@ -462,7 +464,7 @@ extern "C" JNIEXPORT void JNICALL Java_io_github_humbleui_jwm_WindowWayland__1nS
 
     const jchar* bytes = env->GetStringChars(title, nullptr);
     jsize length = env->GetStringLength(title);
-    std::u16string thingie = {reinterpret_cast<const char16_t*>(bytes), length};
+    std::u16string thingie = {reinterpret_cast<const char16_t*>(bytes), static_cast<size_t>(length)};
 
     std::string titleS = std::wstring_convert<
             std::codecvt_utf8_utf16<char16_t>, char16_t>{}.to_bytes(thingie);

```
</details>

dzaima:
glfw after `libdecor_frame_map` has `wl_display_roundtrip`, which seems to be exactly the thing for my hypothesis of wrong event order: https://github.com/glfw/glfw/blob/46cebb5081820418f2a20f3e90b07f9b1bd44b42/src/wl_window.c#L803-L804; things seem to be running reasonably smoothly with `libdecor_decorate; libdecor_frame_map; wl_display_roundtrip(_windowManager.display); libdecor_dispatch;`

dzaima:
Adding a `wl_display_roundtrip(self->_windowManager.display);` after `decorFrameCommit`'s `wl_surface_commit` also improves things?

TheDrawingCoder-Gamer:
For me adding a dispatch after roundtrip, it stalls

TheDrawingCoder-Gamer:
Frames aren't really drawn while paused, unlike on X11, which is odd. Not sure if this is an incorrectness with X11 or Wayland ( I assume wayland, as it only doesn't draw on Raster.) Again, raster is god awful slow ("swapBuffers" creates a new ShmPool and Buffer, which involves memory mapping a shared object. This is called every frame.) I'm capping at about 11 fps on my laptop, compared to max framerate of EGL. 

I considered adding a frame callback, but that just causes the window to never show. I seriously think _some_ form of frame throttling needs to take place bc of how god awful slow this stuff is. Of course down stream in HumbleUI it will never end up being an issue, but it's important for it to at least function. 

Switching layers still doesn't work - I even tried doing a hack by closing and reopening, but it didn't work.

TheDrawingCoder-Gamer:
just realized that events are never actually sent - how interesting

TheDrawingCoder-Gamer:
can't for the life of me figure out why mouse cursor setting isn't working - only the first three (arrow, crosshair, help) load? and text sometimes works???? but I can tell that the hotspot is working??? I hate this

TheDrawingCoder-Gamer:
Rebasing is causing issues on sway. God knows why. 
<details>
<summary> git merge result</summary>
``` 
Auto-merging script/build_utils.py
Auto-merging x11/cc/KeyX11.cc
Auto-merging x11/cc/ScreenInfo.cc
Auto-merging x11/cc/WindowX11.cc
Merge made by the 'ort' strategy.
 .github/workflows/build-deploy.yml          |  1 +
 .github/workflows/rebuild-docker-image.yaml |  2 +-
 CHANGELOG.md                                | 13 +++++++++++
 script/build_utils.py                       |  3 ++-
 shared/cc/impl/JNILocal.hh                  |  2 +-
 x11/cc/KeyX11.cc                            | 99 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++--------------------
 x11/cc/KeyX11.hh                            |  3 ++-
 x11/cc/WindowManagerX11.cc                  | 14 ++++++++----
 x11/cc/WindowX11.cc                         | 11 +++++++---
 9 files changed, 113 insertions(+), 35 deletions(-)
```
</details>
Something is obviously wrong bc none of the files touched are related to wayland in any way. How awful

TheDrawingCoder-Gamer:
nvm it works. There SHOULD:tm: be no more conflict (even though github is being wrong :( )
A flat copy will work (once ready)

TheDrawingCoder-Gamer:
note; you'll need the patch version of skija (pretty sure master branch has the patch, but you have to pull down master and build it). For me it mostly JUST WORKS(TM) on wayland, even with hidpi.

Some jank: 
Positioning is completely impossible. Too bad.
Resizing works only while invisible. Wayland doesn't let you resize yourself non-interactively while mapped.
Resizing ALSO is tied to scale. The dimensions must be divisible by the scale, otherwise it returns early (it's implemented as a bool method but this is never exposed in java).
Vsync is disabled in general. This SHOULD(TM) be ok for both GL and Raster, as the compositor is what actually handles drawing, and it's syncronized properly. the reason vsync is disabled client-side is because under Sway it blocks the whole application if even 1 window is invisible.

There are a lot of weird function changes you can revert, but iirc I didn't touch X11 at all.

It should JUST WORK(tm) on other platforms bc nothing new changed. 

Wayland and X11 work side by side - Wayland is loaded if the envvar `WAYLAND_DISPLAY` is set, and x11 otherwise. There isn't really any error checking here (nor with a lot of this)

Yes, I know code organization is GOD AWFUL. I hate it too. Not used to C++ so bear with this god awful code organization.

I think (?) I've stripped all the debug stuff but I'll do another pass-over soon.

TheDrawingCoder-Gamer:
edge cases that I totally expect to break things
multiple seats (I've never actually seen/used this in the wild, but it's technically part of the spec so I GUESS we should support it)
Window being halfway  between monitors (I don't expect a full on crash here but I expect scaling to not be correct, esp. when moving from a HiDPI screen to a LoDPI screen)

Otherwise I haven't rigorously tested things, only doing stuff that I saw was outright broken. The test itself works fine.

Rn I need to do animated cursors. Well, idk if I _need_ to, but it would be nice to have. My "wait" mouse being still is very jarring to me. 

but rn it's time for eepy : ) have fun reviewing this god awful code in the mean time

TheDrawingCoder-Gamer:
resizing interactively is also a bit jank, likely because the buffer is getting updated way too fast.

Keyboard support is also jank, there are some weird things not supported (notably the compose key, and key repeating).

Going to work on key repeating and maybe compose key? Resizing is also probably more important

TheDrawingCoder-Gamer:
resizing is now _responsive_ on sway and weston, but on weston it sometimes puts up a black screen. Fully functions under wlroots though (tested on sway & wayfire)

TheDrawingCoder-Gamer:
now works without having to use `--skija-dir`. Repeating keys also half functions (technically functions but lags in an odd way). Compose key fully (?) functions, but not at parity with X11. 

For some file conflicts it only lags bc I open the file and use `:wq` instead of `:q!`. `:wq` adds a newline at the end of the file. 

Resizing still jank on weston, but works well enough on wlroots. Need to test on other major compositors (mutter and kwin). It functions by forcing a draw by dispatching a frame event then blocking on the processEvents thing. 

It only blocks if the new screen scale is different from old screen scale, because wayland can crash if you set preferred buffer size to e.g. 2 while the dimensions of the buffer aren't multiples of the scale. Previously this was solved by making it a frame callback but that caused jank.

TheDrawingCoder-Gamer:
...? it doesn't accept any events while window isn't focused. Bad for clipboard bc of how wayland has clipboard; event loop HAS to be running for clipboard copying to work. No idea what/if anything is blocking. This will cause wl-paste to freeze if i try to paste from it, but it pastes if I focus back. Firefox flat-out doesn't paste at all (likely an attempt to prevent stalling bc of pasting). Painful... No idea what's blocking

TheDrawingCoder-Gamer:
it was GL. apparently `eglMakeCurrent` resets swap interval. how upsetting. 

only real thing is animated cursors but that is going to be so so painful.

TheDrawingCoder-Gamer:
please try testing it now - it works enough for me.
weston flickers while resizing on GL, but who cares about weston. Test Mutter (gnome) and KWin (KDE). 

animated cursor support is basically the last thing I need? again I don't really think it's needed but it will make me endlessly upset if I don't add it.

dzaima:
`App.getPrimaryScreen()` appears to still error, which breaks my JWM usage; replacing that with a hard-coded rectangle gets it running.

Scroll wheel is inverted & way too slow for me on Weston, compared to X11 (speed could be a difference between weston config & my native system, but direction shouldn't be).

FPS seems to be uncapped if every frame requests another frame (e.g. in the `./script/run.py` window clicking ctrl+p; also seems that that ctrl+p is sent twice?).

Interacting with the decorations is somewhat buggy, though I don't know how much of that is the fault of whatever outdated versions of libdecor/weston I have from apt.

There are some segfaults around closing windows, which I might look into.

TheDrawingCoder-Gamer:
perhaps I am just not mentally processing everything I have to for C++. I have never used RAII before outside of rust and rust makes it blatantly obvious. I would wager a guess that a vast majority of these are related to RAII in some way.

TheDrawingCoder-Gamer:
In wayland, you are expected to assume that all keys are released, but it's unspecified what you do on enter, when it gives you the keys in an array. I currently process it like they were just pressed, and I assume that's why there is jank on new windows? Very frustrating : / really wish I was writing in either a GC'd language or rust.

dzaima:
RAII only applies to the C++ JWM-specific parts, which I think work fine? Segfaults usually come from within the wayland library, which has its own memory management rules (and cannot be RAII as the interface is all C pointers)

TheDrawingCoder-Gamer:
hmmmm.... does this mean I'm doing something wrong? wonder where

TheDrawingCoder-Gamer:
yeah I kind of feel like I'm out of my depth with debugging C++. I don't write systems languages outside of rust, and rust practically holds my hand compared to C++. I'm going to take a break for the holidays and rotate what I can do now. 

Reverting to draft as well.

TheDrawingCoder-Gamer:
as expected nvidia EGL doesn't work. How shocking. 

it's just a black window with proprietary drivers. Raster works. Will try to fix this, but sway itself requires me to pass `--unsupported-gpu` bc of how god awful the support is.

TheDrawingCoder-Gamer:
have a suspicion that there is memory corruption at play somewhere in pointerLeave and mouseUpdateUnscaled. jank

TheDrawingCoder-Gamer:
somehow managed to break it JUST on non-nvidia sway. wayfire works?

TheDrawingCoder-Gamer:
Have to memory leak frame to prevent crash on close. Jank. This is a bug with the free code of libdecor. 

Weston is so buggy on NVIDIA that I don't really think it's worth trying to test it. Sway and Wayfire work so I assume most wlroots work. 

I've observed that hooking this up to my local UI toolkit causes the window to be black until an update that forces a redraw is sent.

HumbleUI itself seems to be working with the obvious tweaks to prevent breakage with matching on Platform. 

there are some nvidia and mesa exclusive bugs too so it's good I have a laptop and desktop. Getting segfaults on resizing with nvidia that aren't observable in mesa. I think (?) mesa is mostly working but idk. try testing it again to see how busted it is.

TheDrawingCoder-Gamer:
tbh i'm okay with nvidia bugs just cause like... if you are on nvidia and using wayland you are doing something wrong. I'm sure that for older cards the open source drivers work, but propreitary drivers are almost universally unsupported. I'll try to fix it but it's not as high priority as it would be on mesa.

TheDrawingCoder-Gamer:
re: lock cursor. Should I be receiving mouse motion events on mouse move? if so then locking the cursor won't work, doesn't actually send any events. X11 has no impl so my local machine has no reference.

TheDrawingCoder-Gamer:
want to say that winit has a bunch of options in certain places, including primary screens. I think it's ok to just add `Nullable` to getPrimaryScreen and let the IDE handle it. This will likely break downstream projects (including yours) but I'm ok with that. It's better to say there is no data that makes sense then give data that doesn't make sense.

TheDrawingCoder-Gamer:
would like to add touch support but have no touch screen, and afaik my tablet is just a mouse in non supported apps. I _can_ add touch support for my trackpad though.

TheDrawingCoder-Gamer:
weston support is blocked by https://gitlab.freedesktop.org/libdecor/libdecor/-/issues/59. any WM that has decorations will cause errors as GTK doesn't work, and I can't figure out how to force Cairo instead of GTK. would be fixed when the related PR up there is merged, or if I figure out how to force Cairo, as my desktop doesn't use GTK and it's functioning better

TheDrawingCoder-Gamer:
@tonsky it's unhelpful to have this running every time and wasting runtime. idk if you can disable it but please do. I've made local edits to Dockerfile but idk how to really test them.

TheDrawingCoder-Gamer:
not getting anymore crashes related to window closing. Going to remark as ready for review - note for the title bar it no longer uses libdecor and there is no more text - text would require pulling in an extra drawing library on the native level, like cairo, which I'm unsure if wanted. If this is ok point me at a library to use and i'll make it look somewhat nicer. Right now it uses subsurfaces to build a window border for resizing and movement, and subsurface buttons for minimize close and maximize/restore. 

probably still has issues under nvidia, too bad! will test later but as said previously i'm fine with crashing under jank circumstances on wayland for nvidia, bc they are propreitary drivers that can't be debugged.

Notes on vsync: can't really do it (?) Tried delaying the redraw with surface. Causes window not to show, and with extra tinkering to add a force redraw flag, it shows but under sway it doesn't update. Raster will never screen tear as wayland compositors are the ones that actually render out frames. GL _may_ cause screen tearing because the underlying method hooks into the tearing control protocol. This means KWin and Weston may have screen tearing (other compositors don't implement this protocol). All in all I think it's working well enough on Mesa that I would use it downstream. Of course this requires changes downstream in HumbleUI (I already have a local working build).

dzaima:
Wanted to try this out on the just-released Linux Mint 21.3 experimental wayland, but with on latest commit with `sudo apt-get install wayland-protocols wayland-scanner++ extra-cmake-modules` I get this:
```
-- Java home found automatically at /usr/lib/jvm/java-1.11.0-openjdk-amd64. Set JAVA_HOME environment variable to override.
-- Configuring done
-- Generating done
-- Build files have been written to: /mnt/linux2/git/jwm-wayland/x11/build
[7/7] Linking CXX shared library libjwm_x64.so
-- Could NOT find WaylandProtocols: Found unsuitable version "1.25", but required is at least "1.32" (found //usr/share/wayland-protocols)
CMake Error at CMakeLists.txt:15 (message):
  No protocols installed


-- Configuring incomplete, errors occurred!
```
Changing that `1.32` to `1.25` then gives this:
```
-- Java home found automatically at /usr/lib/jvm/java-1.11.0-openjdk-amd64. Set JAVA_HOME environment variable to override.
-- Configuring done
-- Generating done
-- Build files have been written to: /mnt/linux2/git/jwm-wayland/x11/build
ninja: no work to do.
-- Found WaylandProtocols: //usr/share/wayland-protocols (found suitable version "1.25", minimum required is "1.25")
-- Java home found automatically at /usr/lib/jvm/java-1.11.0-openjdk-amd64. Set JAVA_HOME environment variable to override.
-- Configuring done
CMake Error at CMakeLists.txt:38 (add_library):
  No SOURCES given to target: protocols


CMake Generate step failed.  Build files cannot be regenerated correctly.
```


switching to a random old commit of `1d7fb1af4fed7bbd0f40b23569a7b1d86bd59a9c` and building there compiled, but running then crashed Cinnamon, killing all apps, including the browser, making me rewrite this comment :/

TheDrawingCoder-Gamer:
try latest commit - reduces version requirement + uses a var for protocols instead of directly appending.

dzaima:
Fails to build with
```
FAILED: CMakeFiles/jwm.dir/cc/Decoration.cc.o
/usr/bin/c++ -Djwm_EXPORTS -I/mnt/linux2/git/jwm-wayland/wayland/../shared/cc -I/mnt/linux2/git/jwm-wayland/wayland/../linux/cc -I/usr/lib/jvm/java-1.11.0-openjdk-amd64/include -I/usr/lib/jvm/java-1.11.0-openjdk-amd64/include/linux -I/mnt/linux2/git/jwm-wayland/wayland/build -O3 -DNDEBUG -fPIC -std=gnu++14 -MD -MT CMakeFiles/jwm.dir/cc/Decoration.cc.o -MF CMakeFiles/jwm.dir/cc/Decoration.cc.o.d -o CMakeFiles/jwm.dir/cc/Decoration.cc.o -c /mnt/linux2/git/jwm-wayland/wayland/cc/Decoration.cc
/mnt/linux2/git/jwm-wayland/wayland/cc/Decoration.cc:196:6: error: field designator 'wm_capabilities' does not refer to any field in type 'xdg_toplevel_listener'
  196 |     .wm_capabilities = _xdgToplevelWmCapabilities
      |     ~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```
commenting out that line results in it running though! Doesn't look like that Mint's wayland cinnamon is anywhere near complete enough to test on though, weston manages to be less janky.

TheDrawingCoder-Gamer:
bc of how I bind xdg_shell it's ok to omit these versions. Compositors don't send events to clients that bind to lower versions.

TheDrawingCoder-Gamer:
update:
i'm probably not going to work on this anymore. anyone else is free to continue this.
It works well enough for me but I don't really know what else to do here. I think (?) it's good enough but needs more testing

TheDrawingCoder-Gamer:
I took a peek at this again and I think I have an idea of why resizing on sway causes issues: There's a feedback loop in LayerGLWayland.cc as it calls `wl_egl_resize_window` which may cause a xdg_configure, which will force the window to resize... which will end up calling `LayerGL::resize` again.

No clue how to stop this and I am not particularly interested in fixing this myself.

TheDrawingCoder-Gamer:
It's worth noting that this behavior is only bugged while the app is "paused" - it updates fast enough if you force it.
Obviously this will still be a bug for HumbleUI or any sensible UI app that tries to minimize render updates.

EDIT: Nevermind, it's bugged anytime the FPS is low enough.

PR #279, opened 2023-12-03
X11: add KeyLocation, additional keys

dzaima:
Adds KeyLocation for right keys & keypad, and also adds additional Latin 1 keys mapping to shift+another key (not sure if all are strictly necessary, but it can't hurt; started by someone on a Faroese layout (`setxkbmap fo`) hitting `XK_plus` and I just added the rest from the latin-1 section).

The "shift+" part isn't particularly nice as then it's impossible to determine whether the actual shift key is pressed, but it's better than nothing.


tonsky:
Thanks!

PR #284, opened 2024-01-10
fix resize frame

TheDrawingCoder-Gamer:
Resizing doesn't correctly request a new frame. This changes the code so it goes through the platform specific way of requesting a new frame, instead of just accepting a frame event. Fixes resizing oddities in X11 (and wayland when the patch is applied to my branch).

### DIFF
--- a/shared/java/Window.java
+++ b/shared/java/Window.java
@@ -423,7 +423,7 @@ public abstract class Window extends RefCounted implements Consumer<Event> {
         if (e instanceof EventWindowScreenChange) {
             accept(new EventWindowResize(this));
         } else if (e instanceof EventWindowResize && Platform.CURRENT != Platform.X11) {
-            accept(EventFrame.INSTANCE);
+            requestFrame();
         }
     }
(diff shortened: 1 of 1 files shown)


tonsky:
There’s `Platform.CURRENT != Platform.X11`, how can it possibly fix oddities in X11?

TheDrawingCoder-Gamer:
I was mistaken. Works for Wayland at least, fixes oddities there. Before this change, my wayland branch with HumbleUI didn't update the buffer size until I did something that would trigger a repaint anyway, like hovering over a button. Now it resizes it correctly.

tonsky:
How does it help though? Is there some extra work that’s been done in requestFrame? Because you are supposed to reset surfaces in EventFrame hanlder, right? And it shouldn’t matter if it was called from resize event or frame event? It’s called either way?

TheDrawingCoder-Gamer:
I can't observe the changes anymore. Probably a quirk of my system; closing

PR #286, opened 2024-01-19
Add Windows implementation of setTitlebarVisible (#75)

Quezion:
Opening as Windows impl of #75. First commit has working `setTitlebarVisible` in JWM's demo.

I was going to implement more methods but `setTitle(title)` is already implemented & works on my machine. The rest of the methods Tonsky mentions in https://github.com/HumbleUI/JWM/issues/75#issuecomment-1009146537 are Mac specific & may not have clear Windows analogues.

For my own usage I only hide the titlebar. I'm happy to leave Windows style extension to a future PR, but thoughts @tonsky ?

### TODOs
- Additional methods? (might be left to future WinStyle PR)
- [X] Request review from https://github.com/littleli once PR is complete
- [ ] Correctly handle DPI calculation when resizing
- [ ] Refactor out `windowShadow` state tracking variables 

**Bugs**
- [X] Window contents are blurry after resize
  - HumbleUI is not recognizing increase in app pane's draw area that occurs after removing titlebar
  - Need to dispatch or process event in HumbleUI to update render resolution
	  Can probably solve this by handling WM_STYLECHANGED event in `jwm::WindowWin32::processEvent` switch
	  see https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-stylechanged
  - Fixed by running `setContentSize` in `WM_STYLECHANGED`, but has logical side-effect of growing the window by the titlebar-height on every other toggle
    - Should we try to calculate the Windows titlebar height & remove it from the content size when the titlebar is restored? @tonsky 
	  
### Test Instructions
Assuming you have [Windows local build setup](https://github.com/HumbleUI/JWM/blob/main/docs/windows/build.md) & you're in the VS x64 command prompt:
1) `python script/build.py && python script/run.py`
2) Input keyboard shortcut `Ctrl+T` in JWM's demo window to `Toggle Titlebar`. The titlebar should disappear from the top of the window.
3) Run `Toggle Titlebar` again & the default titlebar should reappear

Note that I've only tested this on Windows 10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,8 @@ Alpha. Expect API breakages.
 | setMinimumSize      | ❌   | ❌   | ❌   |
 | setMaximumSize      | ❌   | ❌   | ❌   |
 | setResizable        | ❌   | ❌   | ❌   |
+| bringToFront        | ✅   | ❌   | ❌   |
+| isFront             | ✅   | ✅   | ❌   |
 
 ### Events
 
@@ -224,6 +226,35 @@ Run examples without building (use version from the table above):
 
 ```
 ./script/run.py --jwm-version <version>
+```
+
+### Local JAR
+
+Generate & install a local .jar file:
+
+```
+./script/install.py
+```
+
+This outputs `target/jwm-0.0.0-SNAPSHOT.jar` for use in testing (e.g. `io.github.humbleui/jwm {:local/root "..."}` if using deps.edn)
+
+### MacOS
+
+Before running the build, ensure you've installed:
+* XCode Developer Tools (`xcode-select --install`)
+* Ninja (`brew install ninja`)
+* Python 3 (`brew install python`)
+
+### Debugging
+
+Set `JWM_VERBOSE` in process env to see extra log output when running locally.
+
+``` bash
+# Mac / Linux
+export JWM_VERBOSE=true
+
+# Windows
+set JWM_VERBOSE=true
 ```
 
 # Contributing
--- a/docs/windows/build.md
+++ b/docs/windows/build.md
@@ -25,6 +25,9 @@ or from chocolatey
 ```sh
 choco install -y python
 ```
+
+You'll need to ensure your Java installation's `$JAVA_HOME/bin` is on your system PATH. This exposes necessary interop codefiles like `jni.h` to the compiler. Your JAVA bin path will look similar to "`C:\Program Files\Java\jdk-17.0.2\bin`"
+
 ## Install Ninja
 
 Download executable from https://github.com/ninja-build/ninja/releases and export path.
--- a/examples/dashboard/java/PanelScreens.java
+++ b/examples/dashboard/java/PanelScreens.java
@@ -20,6 +20,8 @@ public class PanelScreens extends Panel {
             titleStyles = new Options("Default", "Hidden", "Transparent", "Unified", "Unified Compact", "Unified Transparent", "Unified Compact Transparent");
         } else if (Platform.X11 == Platform.CURRENT) {
             titleStyles = new Options("Default", "Hidden");
+        } else if (Platform.WINDOWS == Platform.CURRENT) {
+            titleStyles = new Options("Default", "Hidden");
         }
     }
 
@@ -60,6 +62,14 @@ public class PanelScreens extends Panel {
             }
         } else if (Platform.X11 == Platform.CURRENT) {
             WindowX11 w = (WindowX11) window;
+            switch (style) {
+            case "Default" ->
+                w.setTitlebarVisible(true);
+            case "Hidden" ->
+                w.setTitlebarVisible(false);
+            }
+        } else if (Platform.WINDOWS == Platform.CURRENT) {
+            WindowWin32 w = (WindowWin32) window;
             switch (style) {
             case "Default" ->
                 w.setTitlebarVisible(true);
--- a/windows/cc/ThemeWin32.cc
+++ b/windows/cc/ThemeWin32.cc
@@ -18,7 +18,7 @@ bool _isHighContrast() {
         return false;
     }
     bool result = (HCF_HIGHCONTRASTON & highContrast.dwFlags) == 1;
-    JWM_VERBOSE("is HighContrast? '" << result << "'");
+    // JWM_VERBOSE("is HighContrast? '" << result << "'");
     return result;
 }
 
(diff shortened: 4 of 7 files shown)


littleli:
It took me quite a while to successfully setup the dev environment and compile the project. Now I can confirm that after `Ctrl+T` Toggle title bar function hides (and show) window title bar.

window without titlebar

This works both in normal application mode and full screen mode (after `Ctrl+F`). I can also confirm that size of the window after toggle twice increase the window size with the size of title bar. I recommend to consider compensate height as suggested by @Quezion.

Quezion:
Fixed the "window size increasing" behavior, updated `windows.md` with some findings from @littleli , & force pushed a single latest commit to this PR.

I ended up deleting handler in `processEvent WM_STYLECHANGED` and instead doing resize inside of `setTitlebarVisible` --
the problem is that you _must_ know whether the titlebar was added or removed.
* If titlebar was removed, you must resize window to current content size
* If titlebar was added, you must resize content to old window size

**Test Instructions**
1. Execute original test instructions but with repeated `Toggle Titlebar` via `Ctrl+T`
2. Check that the content size should remained fixed (1264, 681) & rendering doesn't degrade into "blurry mode"

Hopefully this is the mergeable commit! 🤞

tonsky:
This does work better! I do think that correcting the size right in the setTitlebarVisible is ultimately the right move.

Can I ask you to change one more thing, though?

The way I feel this should work is that 

```
setWindowSize(w, h)
setTitlebarVisible(false)
```

and

```
setTitlebarVisible(false)
setWindowSize(w, h)
```

should produce the same result. In other words, you tell your window to occupy certain amount of space on a screen. For example, 1/2 on the left. Then, no matter whether you turn titlebar on or off, the amount of space it occupies should stay the same. The external border of the window should not move. Content area will expand/compact a little when removing titlebar, but that’s about it.

I feel this is the most natural behavior. What do you think?

If you agree, then the order you have your conditions is reversed: e.g. if I _remove_ titlebar, I should take _window_ size, hide titlebar, and then set _content size_ to previous coordinates/dimensions.

I tried it myself and it didn’t produce desirable results. Looks like this is because we actually measure window boundary wrong! See #289 

So this is what I suggest. Can you fix #289 first and then we come back here and swap the windowSize/contentSize in the setTitlebarVisible and it will all work flawlessly?

tonsky:
I think remembering is fine? They don’t change anyway, right?

Quezion:
It shouldn't change unless the user changes the Windows theming (removing drop shadow), but that seems like a fine compromise for not having the intermediate possibly-bugged state while we wait for a `WM_` event callback to fire.

Quezion:
I was able to workaround the bugged local behavior but the code is nasty.

Turns out that `setContentSize` uses winapi `AdjustWindowRectEx` -- this _also_ reports incorrect sizes when the titlebar has just been added/removed and the window hasn't received a new render from Windows.

To work around this, we also need to manually track the window's titlebar + borderwidth between hide/set titlebar. The "cleaner" solution would involve some state machine that waits for Window's first render, thus allowing `AdjustWindowRectEx` to actually work, but this introduces other types of complexities.

Quezion:
I've cleaned up the code (as much as is possible) but I need to solve "Windows version detection" before I can submit it. You're supposed to use the below snippet to get the bounds of the application _without_ the dropshadow, but it's only available on Vista or later (majorVersion > 6)

```
DwmGetWindowAttribute(_hWnd, DWMWA_EXTENDED_FRAME_BOUNDS, &rect, sizeof(rect));
```

It turns out that detecting the OS version is nontrivial, the methods have been deprecated multiple times, and Windows would prefer you test for the presence of _features._ Except there's no supported feature that tells you if DwmGetWindowAttributes exists or if you're on Vista+.

https://learn.microsoft.com/en-us/windows/win32/sysinfo/targeting-your-application-at-windows-8-1

So I'm working on a shim to determine OS version for the sake of `getWindowRect()` & will submit the PR once this is complete.

tonsky:
That’s fine I think, we probably won’t run on Vista any time soon. Maybe we should care about Win 10 but not lowerOn 26 Feb 2024, at 20:42, Quest Yarbrough ***@***.***> wrote:﻿
I've cleaned up the code (as much as is possible) but I need to solve "Windows version detection" before I can submit it. You're supposed to use the below snippet to get the bounds of the application without the dropshadow, but it's only available on Vista or later (majorVersion > 6)
DwmGetWindowAttribute(_hWnd, DWMWA_EXTENDED_FRAME_BOUNDS, &rect, sizeof(rect));

It turns out that detecting the OS version is nontrivial, the methods have been deprecated multiple times, and Windows would prefer you test for the presence of features. Except there's no supported feature that tells you if DwmGetWindowAttributes exists or if you're on Vista+.
https://learn.microsoft.com/en-us/windows/win32/sysinfo/targeting-your-application-at-windows-8-1
So I'm working on a shim to determine OS version for the sake of getWindowRect() & will submit the PR once this is complete.

Quezion:
Added latest commit. I misspoke earlier -- this should work on Vista but won't on Windows XP.

The code seems nasty to me. I suspect @tonsky will have some comments to address before the final commit. For the sake of diff, I've just added a 2nd commit instead of squashing both commits down.

tonsky:
Works better, but still not quite there yet. Hiding titlebar keeps window in place (great!) but showing title bar moves window 5 px up (on 200% scaling). If I change scaling to 100%, it only moves 3px up. Is there a DPI-aware calculation lost here somewhere? Perhaps this bit is important?

Screenshot 2024-03-03 at 22 21 21

tonsky:
Re: code itself. I see you change _windowShadowWidth/Height back and forth. Wouldn’t it be better if you just identify them after window is first shown and remember these values?

As I understand, you need to account for shadow if titlebar is visible and not account for it if it’s not. Might this be just a condition in setWindowSize? Something along the lines of

```
if (titlebarVisible) {
  setSize(width + shadowWidth, height + shadowHeight);
} else {
  setSize(width, height);
}
```

I find it easier to work with when state is immutable

Quezion:
I see what you mean with the changes. I'll try to remove the windowShadow vars on next commit & fix the DPI calculation. I remember something in the MS docs about DPI sensitivity & `AdjustWindowRectEx`, which is used in `WindowWin32::SetContentSize()` to calculate some of the window bounds.

I'm sure there's a way to work around it but it'll take some testing to figure it out. I'll figure this out as soon as I can but I'm unexpectedly buying & moving to a new place. It may be a few months before I get my programming time back but I'm looking forward to landing this PR and thanks for the feedback along the way. 🙏

Issue #115, opened 2021-08-19
Mouse scroll amount: px or lines?


tonsky:

jaketothepast:
Hello from `jwind` in #Clojurians slack channel! I'll take this one on as a first issue

jaketothepast:
@tonsky when you say `lines` scroll, does this mean one "tick" of the mouse wheel? I looked into the code for Mac/Windows and it seems to be that way

tonsky:
No, I think in windows there’s a system setting that sets how many “lines” one wheel tick is. We should respect that when reporting it in `EventMouseScroll::deltaLines`. Same for `::deltaPages` (I think it’s also a setting? You’ll have to check)

jaketothepast:
So the "lines" measure is arbitrary on Windows, are we trying to emulate this on X11? I looked into your other commits referencing this issue, seems that Windows is reporting line delta but not pages yet, and the only label on this issue currently is X11

tonsky:
Sorry I might have lost some context since.

So the idea is to report what’s actually happening (e.g. mouse wheel scrolls * lines per wheel tick) as well as some universal unit like pixels. So the people who just want simple solution can use pixels and those will works reasonable on all platforms. But people who want to go extra mile can be faithful to the reality. For example, if user uses mouse, users can add scroll smoothing on top of that, but not if user is on trackpad.

It seems like on Windows it’s already implemented that way.

On macOS it only reports pixels right now, which is how it should be for trackpad, but for mouse it should also report lines. I assume lines are number of ticks of mouse wheel times this setting:

![Screenshot 2024-01-19 at 20 23 28](https://github.com/HumbleUI/JWM/assets/285292/6cea50f3-e792-4bfd-948b-3b3da77da1ef)

As far as I can see, X11 is in the same state as Mac, but I am not sure how realistic it is to get amount of lines from X11.

It also looks like Windows lacks horizontal scroll, which can also be useful. I created a task #285

jaketothepast:
Great, thank you for the additional context, if I have any more questions I'll let you know!

jaketothepast:
I've found how to read that scroll scaling config value:

```objc
- (float)getScrollScaling {
    NSUserDefaults *defaults = [NSUserDefaults standardUserDefaults];
    NSString *scrollScaling = [defaults objectForKey:@"com.apple.scrollwheel.scaling"]; 
    // ...
}
```

[...]

PR #288, opened 2024-01-22
Calculate deltaLines macOS

jaketothepast:
Calculate `deltaLines` from the existing scroll delta. Use the accessibility scaling in calculations.

Related to #115 

Screenshot 2024-01-22 at 8 14 58 AM

### DIFF
--- a/macos/cc/JWMMainView.mm
+++ b/macos/cc/JWMMainView.mm
@@ -3,6 +3,9 @@
 
 #include <algorithm>
 #include <Carbon/Carbon.h>
+#include <CoreFoundation/CoreFoundation.h>
+#include <CoreGraphics/CoreGraphics.h>
+#include <Foundation/Foundation.h>
 #include <jni.h>
 #include "impl/JNILocal.hh"
 #include "impl/Library.hh"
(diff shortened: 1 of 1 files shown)


tonsky:
Sorry, I need to get my hands on a mouse. Will check it out soon!

tonsky:
Okay I did some testing. Some new information:

1. It seems that value of `com.apple.scrollwheel.scaling` is already accounted by macOS. On higher setting, Chrome scrolls much faster, so I guess your app just starts receiving higher deltaY-s
2. Minimal deltaY seems to be 0.1 (one wheel tick), but if you do several one after the other, deltaY starts to get higher. That’s scroll acceleration, and again, you don’t have to do anything, macOS just sends you correct values.
3. By testing both Chrome and Safari, they seem to agree on value of 40 to get from deltaY to px offset. I suggest we do the same

I need you to change some calculations though. Roughly this:

1. Trackpad.
  - Detect by `[event hasPreciseScrollingDeltas] == true`
  - Get EventMouseScroll::deltaX/deltaY from [event scrollingDeltaX]/[event scrollingDeltaY]
  - Do not fill EventMouseScroll::deltaLines, as there is no by-line scrolling on trackpad

2. Mouse
  - Detect by `[event hasPreciseScrollingDeltas] == true`
  - Fill `EventMouseScroll::deltaX/Y` from `[event deltaX/Y] * 40`
  - Fill `EventMouseScroll::deltaLines` from `[event deltaY] * 4` (≈ `[event deltaY] * 40 / CGEventSourceGetPixelsPerLine`)
  
As a side effect, this will let JWM clients distinguish trackpad from the mouse

Let me know if you have better ideas or questions

jaketothepast:
Of course, sorry about that. I'll change the calculations!

jaketothepast:
I promise I haven't forgotten about this! I have a lot of work at my full-time job this week, but when it clears up I will make the changes.

PR #293, opened 2024-09-24
X11: fix _nSetTitle return type

dzaima:
Fixes spurious errors, as the C function has a void return type, thus leaving garbage where JNI expects an object reference. Caused weird NPEs with clang++ 18 on titles longer than 23 chars.


tonsky:
Thank you!

Issue #135, opened 2021-08-25
Set App Icon from Pixels


tonsky:
Follow up to #95 

- [ ] Windows
- [ ] Linux
- [ ] macOS

dzaima:
Currently looking into this, as part of trying to get general app icons for Linux. For that, there are, as far as I can tell, two approaches:

- have a globally-installed `the-app.desktop` file describing the program and its icon location (and the app setting a `WM_CLASS` hint to associate to it); seems impossible to make work via just `setIcon`, but would be good to have as a `WindowX11`-only thing for those wanting to do proper X11 application association;
- set via pixels; this could be bound to `setIcon`, but means something has to decode the png (or whatever file format).

I have both `WM_CLASS` configuring and icon-from-pixels configuring [here](https://github.com/dzaima/JWM/commits/icon/) via a temporary interface, using Skija to decode the png in the dashboard example.

tonsky:
Both look good, make a PR?

I understand that setting icon is more of a packaging problem and probably can’t be easily solved by just doing API calls, but where it _can_ work I don’t see why not have it

PR #294, opened 2024-09-27
X11: setIconData and setClassHint

dzaima:
Implements the X11 part of #135, along with `WM_CLASS` initialization; both only under `WindowX11` currently.


tonsky:
Thanks!

PR #295, opened 2024-09-27
Update X11 info in readme

dzaima:
Only checked X11 things; I'd assume more of the Windows & macOS things should be working too, but leaving that to someone who can check.


tonsky:
Thanks

PR #296, opened 2025-04-10
Fix macos build

SergeevPavel:
`jchar` is an alias to unsigned char. And c++ std doesn't provide a `basic_string` specialisation for it anymore

### DIFF
--- a/shared/cc/StringUTF16.cc
+++ b/shared/cc/StringUTF16.cc
@@ -70,7 +70,7 @@ jwm::StringUTF16::StringUTF16(const uint32_t *str) {
 }
 
 jwm::JNILocal<jstring> jwm::StringUTF16::toJString(JNIEnv* env) const {
-    return jwm::JNILocal<jstring>(env, env->NewString((const jchar*)c_str(), static_cast<jsize>(length())));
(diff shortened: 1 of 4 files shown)


shared/cc/StringUTF16.cc
```diff
@@ -70,14 +70,16 @@ jwm::StringUTF16::StringUTF16(const uint32_t *str) {
...
+    return jwm::JNILocal<jstring>(env, env->NewString((const jchar*)c_str(), static_cast<jsize>(length())));
 }
 
 jwm::StringUTF16 jwm::StringUTF16::makeFromJString(JNIEnv* env, jstring js) {
     jwm::StringUTF16 result;
     jsize length = env->GetStringLength(js);
     const jchar* chars = env->GetStringChars(js, nullptr);
-    result = jwm::StringUTF16(chars, length);
+
+    const char16_t* signed_chars = (const char16_t*)(chars);
```
Alex2772, line 81: Same

```suggestion
    const char16_t* signed_chars = reinterpret_cast<const char16_t*>(chars);
```

tonsky:
Software rot is real! Should work now. Was nice to see you guys again

PR #298, opened 2025-10-19
Update run.py - fix typo that break download

IrrenWirr:
./script/run.py --jwm-version v0.4.18

Breaks, trying to download "https://repo1.maven.org/maven2/io/github/humbleui/jwm/jwm/v0.4.18/jwm-v0.4.18.jar" which doesn't exists.

This commit fix this issue.


tonsky:
Thanks!

Issue #303, opened 2026-03-17
Make press hold and enabled togglable


Folcon:
Hey @tonsky

I've been using HumbleUI to build a game and I just finished wiring up camera panning and realised that `kPressAndHoldEnabled` is on by default
https://github.com/HumbleUI/JWM/blob/dc1b898bd244630029e98e7d13afd56ced965a8f/macos/cc/JWMMainView.mm#L197

This is useful so I'd like to propose adding a toggle so when keyboard input is required I can turn it on and when it's not like when I'm just wanting more raw keyboard input, I can turn it off

Feel free to correct me, I'm not familiar with calling native code from java and I've tried to fit your calling conventions

From what I can see this would involve modifying WindowMac.java and adding a method like:
```java
public static void setPressAndHoldEnabled(boolean enabled) {
    assert _onUIThread() : "Should be run on UI thread";
    _nSetPressAndHoldEnabled(enabled);
}
```

As well as modifying `WindowMac.mm` to add:
```objc
extern "C" JNIEXPORT void JNICALL Java_io_github_humbleui_jwm_WindowMac__1nSetPressAndHoldEnabled
  (JNIEnv *env, jclass cls, jboolean enabled) {
    jwm::kPressAndHoldEnabled = enabled;
}
```

Let me know if you'd prefer me to do it some other way of if you don't think this is a good way to do it =)

Otherwise happy to whip up a PR

From what I can tell I'll also need to do some work with HumbleUI to update it to get it in line with the work you've been doing to update JMW so I can use this, I don't mind putting some time in for that as well

Really want to get back into doing some gamedev on HumbleUI again, your sand example was a good reminder that there's some fun stuff that can be done here!

tonsky:
Yes, all sounds good, please make a PR. It’s quite handy that we already have the flag :)

Folcon:
Hi @tonsky 

I've done a bit more digging, because surprisingly it didn't 100% fix the problem

It looks like it doesn't respect the setting, so JWM is sending to macos to process the events
https://github.com/HumbleUI/JWM/blob/dc1b898bd244630029e98e7d13afd56ced965a8f/macos/cc/JWMMainView.mm#L517-L520

So I tried guarding against that
```objc
// Allow TSM to look at the event and potentially send back NSTextInputClient messages.
if (jwm::kPressAndHoldEnabled) {
    [self interpretKeyEvents:[NSArray arrayWithObject:event]];
}
```
[...]

PR #304, opened 2026-03-20
Added setPressAndHoldEnabled to allow users to toggle that behaviour on the Mac

Folcon:
Just putting this up here and referencing #303

I need to update HumbleUI to make it work with the current JWM to test it properly


Folcon:
Ok, I've updated https://github.com/HumbleUI/HumbleUI/pull/94 as well, this should bring HumbleUI in line with where JWM is and up to date with skija version `0.143.5`

Let me know if you're happy with it or want some changes done to either of these repos =)

PS: Can I also say well done! This is getting awesome! =)

